Issue #51, opened 2024-04-30
Disable-Interactivity failing on Win11 23H2


JMacIV:
https://github.com/Raphire/Win11Debloat/blob/35fc5738f945ef67756003fe8ddea7964a3f039e/Win11Debloat.ps1#L338

Hey, script is great so far! Found a single bug though. 

It seems --disable-interactivity is not available for winget on OneDrive.

Argument name was not recognized for the current command : '--disable-interactivity'

This is Win11 Pro 23H2. OneDrive is still installed :( 

Edited to add that I running with params "-Silent -RemoveApps"

Attaching screenshot of running this command manually. 
![2024-04-30_02-13-43](https://github.com/Raphire/Win11Debloat/assets/43364021/344424f2-fea3-4ac5-97ab-402bcae7e91e)

Raphire:
Heya,

Sorry you're running into this issue. From the provided output it seems like you may be running an old winget version. 

Could you check your winget version for me? This can be checked by running the command: `winget -v`
I have version 1.7.10861 installed for reference.

Thanks!

JMacIV:
Hey Raphire - You are correct, I am running v1.2.10691. This is the default version that comes in 23H2 ISO.

From this page on [Microsoft](https://learn.microsoft.com/en-us/windows/package-manager/), if you click on the versions, it seems they added this "--disable-interactivity" flag with 1.4.

I am fairly certain this PS command will silently install the version from the store... But you might also need to run winget update with that ID afterwards. I could not fully test as this Win11 machine doesn't have a MS store account, but this command got through agreement prompts until interrupted for MS store account.
`winget install -e -h --id=9nblggh4nns1 --source=msstore --accept-package-agreements`

This id value came from this link to the package on MS Store - https://apps.microsoft.com/detail/9nblggh4nns1?rtc=1&hl=en-us&gl=US

Maybe before starting this part of the script, you could add something like 
`$version = (winget -v) -replace 'v', ''
if ($version -lt 1.4) {
    Write-Host "Need to update winget to version 1.4 or later"
    winget install -e -h --id=9nblggh4nns1 --source=msstore --accept-package-agreements
    $version = (winget -v) -replace 'v', ''
    if ($version -lt 1.4) {
        winget upgrade -e -h --id=9nblggh4nns1
    }
}`

However, if you are okay with downloading these straight from the GitHub repos....
`$version = (winget -v) -replace 'v', ''
if ($version -lt 1.4) {
[...]

PR #52, opened 2024-05-03
Winget Version Testing Add

JMacIV:
Added testing for version of winget when setting the "wingetInstalled" var. Updated output to reflect that the version of winget could be outdated.

Related to #51

### DIFF
--- a/Win11Debloat.ps1
+++ b/Win11Debloat.ps1
@@ -331,7 +331,7 @@ function RemoveApps {
         if (($app -eq "Microsoft.OneDrive") -or ($app -eq "Microsoft.Edge")) {
             # Use winget to remove OneDrive and Edge
             if ($global:wingetInstalled -eq $false) {
-                Write-Host "WinGet is not installed, app was not removed" -ForegroundColor Red
+                Write-Host "WinGet is either not installed or is outdated, so $app could not be removed" -ForegroundColor Red
             }
             else {
                 # Uninstall app via winget
@@ -501,6 +501,10 @@ function AwaitKeyToExit {
 # Check if winget is installed
 if (Get-AppxPackage -Name "*Microsoft.DesktopAppInstaller*") {
     $global:wingetInstalled = $true
+    if ((winget -v) -replace 'v','' -lt 1.4) {
+        Write-Output "WinGet is installed, but outdated. Please update to version 1.4 or later."
+        $global:wingetInstalled = $false
+    }
 }
 else {
     $global:wingetInstalled = $false
--- a/Win11Debloat.ps1
+++ b/Win11Debloat.ps1
@@ -498,16 +498,20 @@ function AwaitKeyToExit {
 }
 
 
-# Check if winget is installed
-if (Get-AppxPackage -Name "*Microsoft.DesktopAppInstaller*") {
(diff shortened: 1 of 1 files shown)


Win11Debloat.ps1
```diff
@@ -501,6 +501,10 @@ function AwaitKeyToExit {
 # Check if winget is installed
 if (Get-AppxPackage -Name "*Microsoft.DesktopAppInstaller*") {
     $global:wingetInstalled = $true
+    if ((winget -v) -replace 'v','' -lt 1.4) {
+        Write-Output "WinGet is installed, but outdated. Please update to version 1.4 or later."
+        $global:wingetInstalled = $false
+    }
 }
 else {
     $global:wingetInstalled = $false
```
Raphire:
```suggestion
 # Check if winget is installed & if it is, check if the version is at least v1.4
if ((Get-AppxPackage -Name "*Microsoft.DesktopAppInstaller*") -and ((winget -v) -replace 'v','' -gt 1.4)) {
    $global:wingetInstalled = $true
}
else {
    $global:wingetInstalled = $false

    # Show warning that requires user confirmation, Suppress confirmation if Silent parameter was passed
    if (-not $Silent) {
        Write-Warning "Winget is not installed or outdated. This may prevent Win11Debloat from removing certain apps."
        Write-Output ""
        Write-Output "Press any key to continue anyway..."
        Read-Host | Out-Null
    }
```

Raphire: This output will be overridden immediately by the script menu, so you need to interrupt script execution there with a Read-Host.

Additionally, you can move the version check to the if-statement above it, and show the warning in the else-statement. This way it'll also give a warning if WinGet isn't installed at all 😄.

Raphire (changes requested):
I like the idea of warning users of issues with WinGet. It's not really within the scope of Win11Debloat to update it.

The current warning won't work though, I made a suggestion on how to improve it 😄

Win11Debloat.ps1
JMacIV: This looks like a great suggestion over mine! I figured you would have suggestions, and I would just lay down some base for you to start with :)

JMacIV:
> I like the idea of warning users of issues with WinGet. It's not really within the scope of Win11Debloat to update it.
> 
> The current warning won't work though, I made a suggestion on how to improve it 😄

Suggestion looks good. I have not tested your changes, but it _looks_ good ha. 

I figured you would not want to update it within debloat script. If you did decide to, you could always make it an optional opt-in prompt with a warning that it will either download from MS Store, or GitHub. I will prob maintain this code in my fork since I will want it to automatically install the winget upgrade for my purposes :)

Raphire (approved):

Raphire:
> I figured you would not want to update it within debloat script. If you did decide to, you could always make it an optional opt-in prompt with a warning that it will either download from MS Store, or GitHub. I will prob maintain this code in my fork since I will want it to automatically install the winget upgrade for my purposes :)

I appreciate that for your purposes it's nice to have it install automatically, but I think the warning is enough. I don't really want the script to be responsible for updating stuff like this.

I suspect the next Windows image provided by MS will just come with a more up-to-date WinGet version, so we won't run into this problem anymore. 😄

Issue #55, opened 2024-05-09
feature request: launch debloat script with one liner


loadstring1:
Something similiar to what MAS has. Launch the script with one liner (method 1) without having to download the entire github repo.

example screenshot:
![Zrzut ekranu 2024-05-09 205906](https://github.com/Raphire/Win11Debloat/assets/156520308/c606945d-7d03-4611-aa69-9489412479f1)

loadstring1:
Actually i just made one for you pls review my idea

pull request related to this issue:

https://github.com/Raphire/Win11Debloat/pull/56

loadstring1:
and heres the one liner that i just made it works perfectly tested on windows 11 22h2 home edition

(must be ran with admin perms)

irm https://raw.githubusercontent.com/loadstring1/Win11Debloat/master/get.ps1 | iex

Abyssal-Fox:
I'm not sure if this one liner could be combined with the RemoveAppsCustom list. But I assuming that it may be possible...

Raphire:
Heya, 

This has actually been requested before in issue https://github.com/Raphire/Win11Debloat/issues/39

loadstring1:
> Heya,
> 
> This has actually been requested before in issue #39
> 
> I'll look into this and your PR this weekend when I have more time :)

thank you my script won't support every each windows version because it uses winget to install git then clones your repo and runs "run.bat" which then starts your debloat script

its not the best but works like i wanted it to work

Raphire:
@loadstring1 Thanks for taking the time to work on Win11Debloat. I have adapted your solution provided in #56 to include support for parameters and, among a few other changes, added a check to ensure WinGet is installed. See: 7e89991

The command to run doesn't look as clean, but it'll do the job and avoids using Invoke-Expression (iex). [This article by Microsoft](https://learn.microsoft.com/en-us/powershell/scripting/learn/deep-dives/avoid-using-invoke-expression?view=powershell-7.4) explains why iex should be avoided if possible.

Closing issue as resolved 😎

loadstring1:
> @loadstring1 Thanks for taking the time to work on Win11Debloat. I have adapted your solution provided in #56 to include support for parameters and, among a few other changes, added a check to ensure WinGet is installed. See: [7e89991](https://github.com/Raphire/Win11Debloat/commit/7e899918a8043d85c4b755bcd9f09cc0d527d315)
> 
[...]

PR #56, opened 2024-05-09
Add one liner method of launching windows 11 debloat script.

loadstring1:


Raphire:
Feature has been implemented in 7e89991 based on this solution.

More info here: https://github.com/Raphire/Win11Debloat/issues/55#issuecomment-2120312128

PR #64, opened 2024-05-31
Force remove microsoft edge regardless of region/country.

loadstring1:
Tested and works on my windows 11 home 22h2 it wiped microsoft edge out of my system. I used closed-source programs to help myself with writing this feature.

**Please let me know if i should also consider force uninstalling edge webview and edge update because for now this is just edge uninstaller**

fixes this issue: [38](https://github.com/Raphire/Win11Debloat/issues/38)

PR #66, opened 2024-06-04
Force Remove Edge and OneDrive (without winget)

zoicware:
added two functions to remove edge and onedrive instead of trying to use winget
works on 10 and 11 (latest)

### DIFF
--- a/Win11Debloat.ps1
+++ b/Win11Debloat.ps1
@@ -57,7 +57,7 @@ function ShowAppSelectionForm {
     $global:selectionBoxIndex = -1
 
     # saveButton eventHandler
-    $handler_saveButton_Click= 
+    $handler_saveButton_Click = 
     {
         $global:SelectedApps = $selectionBox.CheckedItems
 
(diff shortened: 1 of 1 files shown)


Raphire:
Heya,

I see little reason to force remove OneDrive like this as the current implementation via WinGet works quite well.

zoicware:
> Heya,
> 
> Thanks for taking the time to contribute 😄
> 
> As I said in #64 I am not a huge fan of force removing applications this way, as it does open the door to many issues if anything changes in the way these apps are installed.
> 
> I do feel like this approach is somewhat justified for edge, considering how nigh impossible it is to uninstall in most regions. However, I see little reason to force remove OneDrive like this as the current implementation via WinGet works quite well. Is there a specific reason to include Onedrive here too?

yep, microsoft insist on further baking edge into the os. I included onedrive since if the user doesnt have or doesnt want winget installed the script sill functions as it should, I dont think it makes sense to do it that way when it can be done natively with powershell but ofc this is your script :)

Raphire:
Closing PR as feature has been implemented in 6049800

PR #119, opened 2024-08-30
Add support for -RemoveAppsCustom and -RunAppConfigurator while using the Quick method

kylehase:
This patch enables the use of `CustomAppsList` within the 'Quick method'. 

Currently, the 'Quick method' dynamically creates its working directory, making it impossible to include `CustomAppsList`, which is expected in the same directory as `Win11Debloat.ps1`. Also, using 'Quick method' to try to create the file using `-RunAppConfigurator` doesn't work since the file is created in the working folder which is deleted.

This update addresses this by:

* Checking for `CustomAppsList` in the directory where the 'Quick method' is initiated and copying it to the temporary folder.
* Checking if `-RunAppConfigurator` was used in the 'Quick method', and if so, moving the `CustomAppsList` file out of the temporary `Win11Debloat` directory to prevent it from being deleted when the script finishes.

### DIFF
--- a/Get.ps1
+++ b/Get.ps1
@@ -75,6 +75,14 @@ Expand-Archive "$env:TEMP/win11debloat-temp.zip" "$env:TEMP/Win11Debloat"
 # Remove archive
 Remove-Item "$env:TEMP/win11debloat-temp.zip"
 
+# Copy CustomAppsList to the Win11Debloat folder if it exists
+if (Test-Path "$PSScriptRoot/CustomAppsList") {
+    Write-Output ""
+    Write-Output "> Found CustomAppsList file"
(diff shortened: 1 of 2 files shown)


Raphire:
Heya,

Thanks for taking the time to contribute! I'll have a look at the changes soon!

Raphire:
I really appreciate the work you put into this PR. I have however opted to implement this in a slightly different manner to keep things a bit more simple.

Both the `CustomAppsList` and `SavedSettings` files are now not removed at all. Considering the files are only a couple KB at most I don't think this is an issue. See: https://github.com/Raphire/Win11Debloat/commit/63a017d620f693b54c319ba18d145fc70d3d1581

kylehase:
Of course. It's your script. I'm just glad the files are now preserved when using the quick method. Thanks.

PR #121, opened 2024-08-31
change the bat file name to Win11Debloat.bat

hst1189:
I think there was an upgrade at once @commit 9d07087. 
However, by changing from 「%~dp0Win11Debloat.ps1」 to 「%~dpn0.ps1」, you can upgrade the version again without making any changes to the content of the file, just by changing the file name.

スクリーンショット 2024-08-31 174005


Raphire:
Heya,

Thanks for taking the time to help contribute to Win11Debloat.

While I understand the benefits of this approach, I also see some issues. 

Because file extensions are hidden by default now (thanks Microsoft) it's likely that people will confuse the bat file and the powershell file if they have the same name. To avoid that I think it's best to keep the filenames distinct.
image

Issue #143, opened 2024-10-15
Extra Debloating


EmperorCrusher:
@Raphire, I use Win11Debloat on all my new PC installs. One thing missing was the ability to debloat Office 365. I personally keep the English version, but debloat the Spanish and French. This also includes removing OneNote. Since I sell Dell PCs to my clients, I also remove a lot of Dell bloatware. I wound up writing a script to facilitate this based on your scripting methods and a LOT of research. If there is any interest in adding any of this to your script, ping me and I'll forward it to you.

I've created another script to do more debloating I have not seen here and can forward it to you as well if you are interested. It includes:

- Removing Microsoft Store from the taskbar (not uninstalling the store, just removing from the taskbar)
- Turn off OneDrive in Settings > Apps > Startup (not removing OneDrive completely)

My script does more, but it's less debloating and more stopping annoyances like making Windows stop asking to make changes after some critical or feature updates. Silence is golden, MS. LMK.

johnnyq:
Super excited for this Please Share, Submit a Pull Request

EmperorCrusher:
I'm not at all experienced at submitting things at Github. I can try following my nose to see if it winds up at the right place or can someone explain how to do this?

Raphire:
Sounds good. I'm always open to adding more options to the script.

> I'm not at all experienced at submitting things at Github. I can try following my nose to see if it winds up at the right place or can someone explain how to do this?

If you fork the repo you can make the changes on your fork and later create a pull request [here](https://github.com/Raphire/Win11Debloat/pulls).

OmaigatTER:
It should be a auto-detection to see if te user uses the app and if it does don't remove it

EmperorCrusher:
I would agree with @OmaigatTER that if this script was to be run on a system **in production**, that it would be best to see if the user actually uses the app. I don't know how to do that, so I'll leave this up to more educated developers like @OmaigatTER to contribute code to help. Would you mind helping?

[...]

PR #146, opened 2024-10-23
DebloatDell Script

EmperorCrusher:
I added a new CMD and PowerShell script called DebloatDell. This deletes Dell's boatware from new Dell PCs except for Dell Update. It also includes code that can be run on any new Windows 11 install to remove Microsoft 365 Spanish and French versions along with OneNote Spanish and French versions. I don't pretend to be an elegant coder. It does the job and that's the hard part.

I also included new.ps1, which has PowerShell code to remove Microsoft Store from the Taskbar and turn off OneDrive from Settings>Apps>Startup. I think it's best for Jeffrey to add it where he wants it into Win11Debloat if he finds it useful. Again, I welcome changes to make it more elegant. It does the job.

### DIFF
--- a/DebloatDell.ps1
+++ b/DebloatDell.ps1
@@ -1,0 +1,164 @@
+# Powershell script to remove:
+
+# uninstall msi packages
+
+Write-Host "Uninstalling MSI Packages..."
+
+Uninstall-package -name "Dell SupportAssist" 
+Uninstall-package -name "Dell Optimizer"
+Uninstall-package -name "Dell Digital Delivery Services"
+Uninstall-package -name "MyDell Application Management"
+Uninstall-package -name "Dell SupportAssist Remediation"
+Uninstall-package -name "Dell SupportAssist OS Recovery Plugin for Dell Update"
+Uninstall-package -name "MyDell Customer Connect"
+Uninstall-package -name "MyDell Components Installer"
+Uninstall-package -name "Dell Core Services"
+
+# uninstall AppxPackages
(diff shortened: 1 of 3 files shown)


OmaigatTER:
Some people use onedrive disabling it from auto start may cause data lose. Also what if the user is Spanish and it uses Microsoft 365.

EmperorCrusher:
Just so this is clear to everyone, I made this fork simply to allow a way for this code to be made available to @Raphire. I fully expect that if Jeffrey does anything with it, he will take the code to turn off OneDrive and remove Microsoft Store from the taskbar and implement it in 11Debloat the way he does now, by making these things optional and allowing the end user to determine whether to execute it or not. This was not provided to make people fret, but to do the heavy lifting by providing the basic code necessary to do the job.

As for concerns of the user being Hispanic, again, this is intended to be added to Win11Debloat as an option to remove. I would not be surprised if Jeffrey took the code and added an option to remove the English version for those friends living in non-English speaking countries. I even found a Brazilian version on a laptop once and found I could alter the code to uninstall the Brazilian version as well (not included in this script). I would therefore imagine that the code could be altered or added to uninstall any language version in the same way as the end user requires.

If it helps you to understand better, I'm a solo IT provider and purchase dozens of PCs/laptops from Dell for my clients. I spent a lot of time doing the debloating of Dell and MS products, so I came up with this script. This works best for me. It would not work best for my counterparts in France or Mexico or Spain. I'm providing it as is to be beneficial to anyone who wants to take it and run with it to make it better or more well suited for their own circumstances.

Raphire:
Closing this PR, I'll leave https://github.com/Raphire/Win11Debloat/issues/143 open to refer to this.

PR #169, opened 2024-12-23
fix: use env:USERNAME

kapad:
Use the env username in the function to delete start menu items. 

I was in a situation where the home folder was auto generated when I installed windows, and I then switched to a local account with a different user name which caused this script to fail.


Raphire:
Nice find, thanks for taking the time to fix this.

Issue #174, opened 2025-01-09
Sysprep option errors when user name is substring of "Users"


Rsge:
**Describe the bug**
The Sysprep option throws an error when the currently logged in user has a user name that is a substring of "Users", e.g. "User".
The error is thrown because the replace function to get the default path replaces both the user name and the substring equalling the user name in "Users", for example:
`C:\Users\User` is converted to `C:\Default\NTUSER.DATs\Default\NTUSER.DAT`.

**To Reproduce**
1. Have an account with a name that is a substring of `Users`.
2. Try running the script with the Sysprep mode.

**Expected behavior**
No error.

**Additional Info**
I know this script will be run on "Administrator" 99.9 % of the time, which is probably why this bug never came up, but still.

Raphire:
Fixed as of 2818854

Thanks @Rsge!

PR #175, opened 2025-01-09
Fix: Use regex replace to avoid #174

Rsge:
Change the function used to replace the result of `$userprofile` to `Default` from `.Replace()` to `-Replace` to enable use of Regex in matching to avoid #174.

### DIFF
--- a/Win11Debloat.ps1
+++ b/Win11Debloat.ps1
@@ -574,7 +574,7 @@ function ReplaceStartMenuForAllUsers {
     }
 
     # Get path to start menu file for all users
-    $userPathString = $env:USERPROFILE.Replace($env:USERNAME, "*\AppData\Local\Packages\Microsoft.Windows.StartMenuExperienceHost_cw5n1h2txyewy\LocalState")
+    $userPathString = $env:USERPROFILE -Replace ('\\' + $env:USERNAME + '$'), "*\AppData\Local\Packages\Microsoft.Windows.StartMenuExperienceHost_cw5n1h2txyewy\LocalState"
     $usersStartMenuPaths = get-childitem -path $userPathString
 
     # Go through all users and replace the start menu file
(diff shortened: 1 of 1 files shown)


Win11Debloat.ps1
```diff
@@ -574,7 +574,7 @@ function ReplaceStartMenuForAllUsers {
     }
 
     # Get path to start menu file for all users
-    $userPathString = $env:USERPROFILE.Replace($env:USERNAME, "*\AppData\Local\Packages\Microsoft.Windows.StartMenuExperienceHost_cw5n1h2txyewy\LocalState")
+    $userPathString = $env:USERPROFILE -Replace ('\\' + $env:USERNAME + '$'), "*\AppData\Local\Packages\Microsoft.Windows.StartMenuExperienceHost_cw5n1h2txyewy\LocalState"
```
Raphire: Thanks for taking the opportunity to fix this yourself. I did find that this one here is missing a \ still. Check my suggestion:

```suggestion
    $userPathString = $env:USERPROFILE -Replace ('\\' + $env:USERNAME + '$'), "\*\AppData\Local\Packages\Microsoft.Windows.StartMenuExperienceHost_cw5n1h2txyewy\LocalState"
```

You can simply run the command in powershell to check the output to confirm.
Before: `C:\Users*\AppData\Local\Packages\Microsoft.Windows.StartMenuExperienceHost_cw5n1h2txyewy\LocalState`

After: `C:\Users\*\AppData\Local\Packages\Microsoft.Windows.StartMenuExperienceHost_cw5n1h2txyewy\LocalState`

Rsge: Wow, I missed like all the things first and made a lot of carelessness mistakes after, which is the reason for the many force pushes – but that one I still missed in the hubbub ^^"

PR #192, opened 2025-02-10
Fixes #191: Skip Explorer restart when in sysprep mode

### LINKED ISSUE #191: Do not restart Explorer when in `sysprep` mode
**Describe the bug**
When in `sysprep` mode, we are only applying changes to the default user profile. There is no need to Restart Explorer in this case. Further, it can hamper cmd script execution.

**Additional context**
I just encoutered this. I wanted to call windebloat with `-silent` and `-sysprep` from another cmd script form OOBE Shift-F10-Console. My script terminates because RestartExplorer kills the CMD window as it is a subprocess from the running Explorer instance.

PR follows.

helge000:
- Since Settings are applied to Default Profile, calling RestartExplorer does nothing
- RestartExplorer also terminates all processes witch might have called WinDebloat in the first place

### DIFF
--- a/Win11Debloat.ps1
+++ b/Win11Debloat.ps1
@@ -545,16 +545,19 @@ function RegImport {
 
 # Restart the Windows Explorer process
 function RestartExplorer {
+    if ($global:Params.ContainsKey("Sysprep")) {
+        return
+    }
+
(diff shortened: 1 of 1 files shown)


Raphire:
Thanks for taking the time. I made some minor adjustments to how this is handled to just jump out of the restart function alltogether when in Sysprep Mode. I think this handles it more gracefully.

PR #212, opened 2025-03-10
Added Disable Fast Startup Option

SimplyYourAverageDev:
Added the option to disable fast startup when running this script.

P.S. I am a very sloppy coder, might have made some mistakes (even tho code has been double checked by Claude 3.7 Sonnet).

Review my changes carefully as I might have forgot things, but I think I got it all right this time.


Raphire:
I've added a few things that were missing, thanks for taking the time to create this PR.

PR #215, opened 2025-03-22
Running script with -Verbose parameter now shows app uninstall progre…

AshiK6907X:
…ss bars

Linting


Raphire:
Heya,

Your branch seems to be a bit behind, as the script should already show the uninstall progress when running with the `-Verbose` parameter. 

https://github.com/Raphire/Win11Debloat/blob/f1e683476ec853b04112a8901908092a8272d37d/Win11Debloat.ps1#L1119-L1127

Issue #231, opened 2025-05-04
Error on creation of sys restore point


lucas-timeworkstudio:
**Describe the bug**
Error in main script, ln 759. I did not go that deep into this so I don't have the actual issue yet but ill try to get it

**To Reproduce**
I do not have the steps, If you need them I can try to remember what options I picked on custom mode.

**Screenshots**

![Image](https://github.com/user-attachments/assets/6cb38d31-84be-4452-bd54-ac246e6f37e5)

**Additional context**
Fresh Install of Windows 11-Pro on VM Box.
Local User

lucas-timeworkstudio:
To save whoever time the exact line is

**Checkpoint-Computer -Description "Restore point created by Win11Debloat" -RestorePointType "MODIFY_SETTINGS"**

and with context is

if ($recentRestorePoints.Count -eq 0) {
        **Checkpoint-Computer -Description "Restore point created by Win11Debloat" -RestorePointType "MODIFY_SETTINGS"**
        Write-Output "System restore point created successfully"
    }
    else {
        Write-Host "A recent restore point already exists, no new restore point was created." -ForegroundColor Yellow
    }

lucas-timeworkstudio:
I believe this comes from how the VM Interacts with the script. However im going to keep this issue open until it gets resolved because I can't 100% say that its the VM

Raphire:
It's possible that the system restore service is disabled on the VM. For now I have updated the script to handle an error when creating a restore point. [a888b9a](https://github.com/Raphire/Win11Debloat/commit/a888b9ae427e7e3bc6e516737e4b3f63f2db27e4)

lucas-timeworkstudio:
> It's possible that the system restore service is disabled on the VM. For now I have updated the script to handle an error when creating a restore point.
> 
> Edit: [d294eb6](https://github.com/Raphire/Win11Debloat/commit/d294eb6b7a3ebff02e041e77c50e61fb856e3cd6)

Just retried with the new script and The same issue persists.

lucas-timeworkstudio:
It's fine for now, but if anyone else needs vm testing this should be patched

Raphire:
> > It's possible that the system restore service is disabled on the VM. For now I have updated the script to handle an error when creating a restore point.
> > Edit: [d294eb6](https://github.com/Raphire/Win11Debloat/commit/d294eb6b7a3ebff02e041e77c50e61fb856e3cd6)
> 
[...]

PR #233, opened 2025-05-04
Check & Stop for VM Restore Point

lucas-timeworkstudio:
Patched #231 

Comparing might look weird because I copied from a code space on the base and pasted into my local fork

### DIFF
--- a/Appslist.txt
+++ b/Appslist.txt
@@ -137,3 +137,23 @@ XING
 #Microsoft.windowscommunicationsapps    # Mail & Calendar
 #Microsoft.XboxGameOverlay              # Game overlay, required/useful for some games
 #Microsoft.XboxGamingOverlay            # Game overlay, required/useful for some games
+#AD2F1837.HPAIExperienceCenter
+#AD2F1837.HPConnectedMusic
+#AD2F1837.HPConnectedPhotopoweredbySnapfish
+#AD2F1837.HPDesktopSupportUtilities
+#AD2F1837.HPEasyClean
+#AD2F1837.HPFileViewer
+#AD2F1837.HPJumpStarts
+#AD2F1837.HPPCHardwareDiagnosticsWindows
+#AD2F1837.HPPowerManager
+#AD2F1837.HPPrinterControl
+#AD2F1837.HPPrivacySettings
+#AD2F1837.HPQuickDrop
+#AD2F1837.HPQuickTouch
+#AD2F1837.HPRegistration
+#AD2F1837.HPSupportAssistant
+#AD2F1837.HPSureShieldAI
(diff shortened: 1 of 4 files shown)


lucas-timeworkstudio:
+ Grouped what the script actually removes should work the same though, Don't know if you want it this way

Raphire:
It doesn't seem to properly detect the fact that system restore is enabled on my system. Is this the correct service?

P.S. I found [this reddit thread here](https://www.reddit.com/r/PowerShell/comments/1dxqmnj/system_restore_rabbit_hole/), suggesting it's possible to use the value of the `HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows NT\CurrentVersion\SystemRestore` registry key, specifically with the `RPSessionInterval` value. (0 being off, >0 being on)

lucas-timeworkstudio:
> It doesn't seem to properly detect the fact that system restore is enabled on my system. Is this the correct service?
> 
> P.S. I found [this reddit thread here](https://www.reddit.com/r/PowerShell/comments/1dxqmnj/system_restore_rabbit_hole/), suggesting it's possible to use the value of the `HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows NT\CurrentVersion\SystemRestore` registry key, specifically with the `RPSessionInterval` value. (0 being off, >0 being on)

Let me do some testing on this and see if it works, Ill get back to you tomorrow. I'll be out for the rest of the day

lucas-timeworkstudio:
> Sounds good. I've updated the PR to include the latest changes from master. That'll make it easier to evaluate the changes.

Hm so I tried adding some stuff, then tested it and it keeps crashing, no logs, no trace of anything. Im going to try doing a few more things but it might come in finished a bit delayed

Raphire:
I've been experimenting with this a bit. Check out my changes here: https://github.com/Raphire/Win11Debloat/compare/master...system-restore

I've made it so a system restore point will always be created in custom mode and removed the prompt. In addition I've added a check for the registry key that indicates whether system restore is enabled. If it isn't it will attempt to enable it. Seems to work pretty well.

lucas-timeworkstudio:
> I've been experimenting with this a bit. Check out my changes here: [master...system-restore](https://github.com/Raphire/Win11Debloat/compare/master...system-restore)
> 
> I've made it so a system restore point will always be created in custom mode and removed the prompt. In addition I've added a check for the registry key that indicates whether system restore is enabled. If it isn't it will attempt to enable it. Seems to work pretty well.

If you want to merge your ill close this one

lucas-timeworkstudio:
> I've merged it (See #235) Thanks for your help with this :D

Your welcome, thank you for assisting me on that

PR #260, opened 2025-06-16
Updated app descriptions

amarcum:
Updated the app descriptions that were not already included in the list. Hopefully it makes it easier to understand which app is removed. No new apps were added and no existing apps were removed, only updated the descriptions of that the apps are.


Raphire:
Heya, Thanks for taking the time to add these descriptions!

PR #280, opened 2025-07-10
Add "Last Active Click" Taskbar Option

cwshugg:
Hiya - this PR adds the ability to enable the "Last Active Click" behavior for the taskbar. It allows you quickly switch focus between multiple active windows from the same application by repeatedly clicking on the application's icon. I've added appropriate regfiles to enable this feature, undo it, and enable it in `sysprep` mode. I've tested this locally on my machine and all seems to be working great.

Here are a few screen recordings showing what I'm referring to:

### Before Enabling 'Last Active Click'

([Link to GIF](https://shugg.dev/images/windows/last_active_click/lastactiveclick_disabled.gif))

### After Enabling 'Last Active Click'

([Link to GIF](https://shugg.dev/images/windows/last_active_click/lastactiveclick_enabled.gif))

Let me know if there's anything I missed or if I can update/improve something. One thing I'm unsure about is the minimum version to enforce for this feature. I know this has been around at least since Windows 10, but how far back, I'm unsure...

### DIFF
--- a/Assets/Menus/Info
+++ b/Assets/Menus/Info
@@ -42,6 +42,7 @@ Taskbar
 - Disable the widgets service & hide icon from the taskbar.
 - Hide the chat (meet now) icon from the taskbar.
 - Enable the 'End Task' option in the taskbar right click menu. (W11 only)
+- Enable the 'Last Active Click' behavior in the taskbar app area.
 
 Start
 - Disable & hide the recommended section in the start menu. (W11 only)


Raphire:
@cwshugg Thanks for taking the time to add this feature. I didn't know about this, but I'm very interested to enable this myself.

As for the PR itself, it looks good, but you missed adding the parameter to Get.ps1 (which passes it through to the main script) and adding the feature to the app info screen. I quickly added those for you. Other than that it looks all good!

cwshugg:
Hey @Raphire, happy to contribute - thanks for the review and catching the spots I missed!

PR #284, opened 2025-07-22
README.md: Fix capitalization

C0rn3j:
The OS is called `macOS`, not `MacOS`


LouisKMills:
cannot imagine a more essential change

Raphire:
Heya, I don't think this is a necessary change.

PR #298, opened 2025-08-20
Properly check if Modern Standby is supported

Raphire:
@SashankBhamidi 

Added a function that actually checks whether Modern Standby is supported on this PC. Also updated the prompt so it's clear this applies to all users.

### DIFF
--- a/Win11Debloat.ps1
+++ b/Win11Debloat.ps1
@@ -542,6 +542,39 @@ function Strip-Progress {
 }
 
 
+# Check if this machine supports S0 Modern Standby power state. Returns true if S0 Modern Standby is supported, false otherwise.
+function CheckModernStandbySupport {
+    $capture = $false
+    
+    try {
+        $null = switch -Regex (powercfg /a) {
+            'The following sleep states are available on this system:' {
+                $capture = $true
+            }
+        
+            '^\s{2,4}(Standby \(S0 Low Power Idle\))' {
+                if($capture){
+                    return $true
+                }
+            }
+        
+            'The following sleep states are not available on this system:' {
+                $capture = $false
+            }
+        }
+    }
+    catch {
+        Write-Host "Error: Unable to check for S0 Modern Standby support, powercfg command failed" -ForegroundColor Red
+        Write-Output ""
+        Write-Output "Press any key to continue..."
+        $null = [System.Console]::ReadKey()
+        return $true
+    }
+
+    return $false
+}
+
+
 # Import & execute regfile
 function RegImport {
     param (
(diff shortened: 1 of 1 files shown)


SashankBhamidi:
@Raphire Good improvement using actual Modern Standby detection instead of battery check.

**Issue**: Language dependency. Function looks for English strings that are localized:
- French: "Les états de veille suivants sont disponibles"  
- German: "verfügbaren Ruhezustände" 
- Will fail on non-English Windows

**Issue**: Error fallback returns `true` when powercfg fails, potentially showing option when detection actually failed.

Still significant improvement over battery-only detection.

SashankBhamidi:
Language-independent implementation suggestion:

```powershell
function CheckModernStandbySupport {
    try {
        $output = powercfg /a
        # Look for S0 identifier instead of localized text
        return ($output -match 'S0.*Low Power Idle|S0.*[Ii]dle')
    }
    catch {
        Write-Host "Error: Unable to check Modern Standby support" -ForegroundColor Red
        return $false  # Default to false when detection fails
    }
}
```

This approach:
- Searches for universal "S0" identifier present in all languages
- Defaults to `false` on detection failure  
- Avoids parsing localized header text

SashankBhamidi:
Additional considerations:

**Modern Standby variants**: Should also handle "Network Connected" vs "Network Disconnected" states - both indicate Modern Standby support.

**Performance**: Consider caching the result in a script variable to avoid repeated `powercfg` calls if the function is used multiple times during execution.

Raphire:
> Language-independent implementation suggestion:
> 
> ```powershell
> function CheckModernStandbySupport {
>     try {
>         $output = powercfg /a
>         # Look for S0 identifier instead of localized text
>         return ($output -match 'S0.*Low Power Idle|S0.*[Ii]dle')
>     }
>     catch {
>         Write-Host "Error: Unable to check Modern Standby support" -ForegroundColor Red
>         return $false  # Default to false when detection fails
>     }
> }
> ```
> 
> This approach:
> 
> * Searches for universal "S0" identifier present in all languages
> * Defaults to `false` on detection failure
> * Avoids parsing localized header text

@SashankBhamidi Good point, it's indeed not language independent which I completely overlooked. Unfortunately your example will not work either, as it will match regardless of whether it's enabled or not.

This is because when not enabled it outputs this:
```
The following sleep states are available on this system:
    Standby (S3)
    Hibernate

The following sleep states are not available on this system:
    Standby (S1)
        The system firmware does not support this standby state.

    Standby (S2)
        The system firmware does not support this standby state.

    Standby (S0 Low Power Idle)
        The system firmware does not support this standby state.

    Hybrid Sleep
        The hypervisor does not support this standby state.

    Fast Startup
        This action is disabled in the current system policy.
```

I have however managed to get something working that seems to be working. Could you try it on your end to confirm?

SashankBhamidi:
Solid approach, @Raphire! 

One minor suggestion - the regex `'(.*S0.{1,}\))'` could be more specific like `'S0.*Low Power Idle'` to avoid potential false matches, though it works fine for the current powercfg output format.

Raphire:
> Solid approach, @Raphire!
> 
> One minor suggestion - the regex `'(.*S0.{1,}\))'` could be more specific like `'S0.*Low Power Idle'` to avoid potential false matches, though it works fine for the current powercfg output format.

Would that not make it language dependant again? In any case, I doubt this will be changed much in the future. If there is any change the impact is minimal.

SashankBhamidi:
@Raphire You're right about language dependency.

But the current regex `'(.*S0.{1,}\))'` is still overly broad - `{1,}` matches any character sequence after S0.

Could tighten to just `'S0.*\)'` - keeps it language-independent while being more precise.

I'm just nitpicking at this point. The functional difference is meaningless for the actual `powercfg /a` output format.

Raphire:
It is a little broad now, I agree. But for now I feel this is fine. If things change in the future I'll be sure to update it ASAP.

PR #321, opened 2025-09-12
add settings file

jlssmt:
Could we please have something like this?
I would like to be able to compare my SavedSettings file with the file in the repository in order to see what has changed or been added.
I would also like to be able to comment/uncomment stuff without running the script.

For apps, this is already working with Applist.txt.

Maybe this file could replace the entire 'parameter' section of the wiki, or at least the parameters from the questionnaire.


Raphire:
Heya,

I do like the idea of this and could see it work as an addition to the existing options of customizing the script's behaviour.

However, a pull-request is not the best way to request this feature. You can create a feature request here: https://github.com/Raphire/Win11Debloat/issues

PR #328, opened 2025-09-15
Fixed LICENSE File section with working link.

blackXploit-404:


Raphire:
Heya, I don't really think this is a necessary change, I also don't know if github still properly recognizes the license file with this change.

PR #329, opened 2025-09-15
Added an option to choose combine mode for Taskbar buttons

tsumi:
Not everyone likes the macOS-style taskbar :)


Raphire:
Heya,

Thanks for taking the time to add this. I think this will be a nice addition to the script. 

I have made a few minor adjustments, mainly changing the wording and adding the new parameter keys to the `Get.ps1` file aswell, so they can be passed on to the main script properly.

Issue #345, opened 2025-09-29
Combine mode for multi-monitor Taskbar


Jenso-DE:
### Checklist

- [x] I have searched for existing issues/discussions and didn't find any similar ones.

### Describe the feature or improvement

Appreciate this latest enhancement: https://github.com/Raphire/Win11Debloat/pull/329#issue-3419050564 💯 @tsumi 
Would love to have the same for multi-monitor. 

And additionally this one:
Show app icons on the taskbar where the window is open (Multiple monitors)

Key: HKEY_CURRENT_USER\Software\Microsoft\Windows\CurrentVersion\Explorer\Advanced

Value (DWORD): MMTaskbarMode

Possible values:

0: Duplicate, 1: Primary and monitor window is on, 2: Monitor window is on.

### Examples

Key: HKEY_CURRENT_USER\Software\Microsoft\Windows\CurrentVersion\Explorer\Advanced

Value (DWORD): MMTaskbarGlomLevel

Possible values:

0 = Always combine, hide labels
1 = Combine only when taskbar is full
2 = Never combine

tsumi:
I don't usually use a multi-monitor setup but the request makes sense.
Should `MMTaskbarGlomLevel` follow the `TaskbarGlomLevel` value or there are valid reasond to keep it separated?

tsumi:
(Meanwhile I implemented a preview with separate options)

Raphire:
Thanks @tsumi for implementing this with #346

PR #346, opened 2025-09-30
Added options to manage taskbar in multi-monitor mode

tsumi:
This want be an enhancement for: https://github.com/Raphire/Win11Debloat/issues/345

- Added option to combine or not icons and labels in taskbar in multi-monitor too
- Added an option to choose which icons show on each taskbar in multi-monitor mode.

**I done some tests and seem works fine, more tests are appreciated.**

TODO:

The option for `Show app icons on the taskbar where the window is open (Multiple monitors)` option must be re-checked on a English Win11 installation (I'm too lazy to install and switch language :P)

In IT localization the options looks like (and act) a little different:

`Key: HKEY_CURRENT_USER\Software\Microsoft\Windows\CurrentVersion\Explorer\Advanced`

Value (DWORD): `MMTaskbarMode`

Possible values:

0 = Show all app icons on ALL taskbars
1 = Show app icons on main taskbar AND on taskbar where the windows is
2 = Show apps icons ONLY on taskbar where the window is

**If someone can check the English translation and/or post a screen I'll update them.**


Jenso-DE:
Can confirm for DE localization like your findings.

Found this in https://learn.microsoft.com/en-us/windows/apps/develop/settings/settings-common for MMTaskbarMode
0: Duplicate, 1: Primary and monitor window is on, 2: Monitor window is on.
So it's the same for all languages.

Will edit my issue with the correct values.

tsumi:
Thanks, updated texts to something more understandable.

Raphire:
Heya,

Thanks again for taking the time to add this. 

I do think that specifically the combine mode for the main and multi-monitor taskbar should be combined into one prompt and change both to the same value. They can be left as separate registry files and parameter so the script can do them separately if you use launch parameters, but the custom mode is getting a bit bloated with more and more prompts being added lately.

Edit: I have made the changes discussed above, and added missing parameter keys to get Get.ps1 file so they're properly passed to the main script when using that launch method.

tsumi:
Makes sense, thanks for the merge!

PR #355, opened 2025-10-07
unattended defaults run

TheKitoInc:
run unattended defaults manual or from autounattend.xml on windows install media


N0tHorizon:
My question is why do you need to make a new file for this and not just incorporate it into the ps script

Raphire:
I agree with @N0tHorizon, in addition, the hardcoded parameters make it very specific for one usecase.

PR #370, opened 2025-11-04
Add NoRestartExplorer switch

oyxnaut:
When running Win11Debloat as a scheduled task triggered by Windows update, the Explorer.exe process can be restarted multiple times a day. Since this causes flickering, and the loss of all opened file explorer windows, a new script parameter `-NoRestartExplorer` has been added.


Raphire:
Looks like a nice addition. I have added the parameter key to the `Get.ps1` file aswell so it correctly passes the parameter when launching the script via the quick method.

PR #383, opened 2025-12-01
Header

Frafou:
updated both PS1 files with header information.


Raphire:
Heya,

I don't really think this is something that is needed. I'd rather keep this kind of stuff in the wiki, instead of having this kind of information in the file as a header.

Issue #366, opened 2025-10-24
OneDrive is not uninstalled 25H2


BigxxRed:
### Checklist

- [x] I have searched for existing issues/discussions and didn't find any similar ones.
- [x] I haven't used any other scripts, tools or programs that might have caused this issue.

### Windows version

Windows 11 25H2 26200.6584 

### Script mode/options

Power shell while in Audit Mode 

### Describe the issue

Thank you for your great tool.
I noticed while Running Debloat  within Audit Mode the script seems to run fine but have noticed that OneDrive is not removed. Even after a restart OneDrive has to manually be uninstalled with separate script. Quick Access disabling also requires manual steps. When opening file explorer OneDrive is in the Quick Access  top portion of file explorer. Is there a way to include disabling of Quick Access?? 

### Steps to reproduce

I have included custom apps list and settings files.
New install from untouched Microsoft 25H2 iso. Entered Audit mode installed Microsoft Office ODT without internet connection.  ODT configuration file included as well.

[configuration.xml](https://github.com/user-attachments/files/23135602/configuration.xml)

[CustomAppsList.txt](https://github.com/user-attachments/files/23135624/CustomAppsList.txt)

[SavedSettings.txt](https://github.com/user-attachments/files/23135628/SavedSettings.txt)

Raphire:
Heya,

Thanks for the detailed report. It seems like Windows will actually install OneDrive after the user logs in, even if it's uninstalled in Audit Mode. I'll see if I can find a good way around this.

Edit: I'll also look into disabling the OneDrive entry in Explorer

BigxxRed:
Thank you for your time. You are greatly appreciated.

BigxxRed:
[script Remove-OneDrive-n-Quick-Access-Folders.txt](https://github.com/user-attachments/files/23143181/script.Remove-OneDrive-n-Quick-Access-Folders.txt)

I have included what is currently working for me after running your script. perhaps something can help.

Raphire:
@BigxxRed As of e66103f. The script will now create a scheduled task the next/first time the user logs in that will uninstall OneDrive. In my testing this seems to work quite consistently. 

Are you in the position to test this? If so, make sure to manually download the latest version from the master branch and run that directly.

BigxxRed:
I will  test today or tomorrow at the very latest. I'll get back to you. Thanks.

Raphire:
[...]

PR #423, opened 2026-01-09
Refactor OneDrive uninstall handling

Jenso-DE:
Until now, Uninstall_Microsoft_OneDrive.reg has not been used. 
Now it is used for -DisableOneDrive.


Raphire:
I appreciate you taking the time to create a PR, however the `-DisableOneDrive` parameter is a legacy parameter that does the same as the `-HideOneDrive` parameter. This feature is only meant to remove the OneDrive entry from File Explorer. As such I don't think it's a good idea to change the behaviour of that parameter so drastically. 

Also, as explained in my comment [here](https://github.com/Raphire/Win11Debloat/issues/366#issuecomment-3751293698) the script does actually use the registry file you mentioned. It's dynamically called when you opt to remove OneDrive in Sysprep mode.

PR #441, opened 2026-02-02
Improve UI accessibility for screen readers

makhlwf:
This PR significantly improves the accessibility of the Win11Debloat UI for users relying on screen readers (such as Windows Narrator or NVDA). By adding AutomationProperties, interactive elements are now properly labeled and announced

Why this is needed:
Previously, many buttons and inputs lacked accessible names, causing screen readers to announce them generically (e.g., just "Button" or "Unlabeled"). This change ensures all interactive elements have descriptive labels, allowing for a fully accessible experience.


Raphire:
Thanks for taking the time to add this, looks like a great addition!

PR #451, opened 2026-02-04
Add option to disable BitLocker automatic device encryption

Raphire:


pgalkin:
What if it's already been encrypted? Will this break decryption?

Raphire:
@pgalkin This only prevent the system from automatically encrypting drives that are not already decrypted. It will not remove encryption from drives already encrypted.

PR #491, opened 2026-02-23
add warning about removing Edge and its impact on Sandbox

AnonymousWP:
The only thing I'm not sure about is if this is the right place to do it, because the description may actually be too long now. I've noticed other descriptions are not this long. Not sure what a better place is for that warning.

### DIFF
--- a/Apps.json
+++ b/Apps.json
@@ -544,7 +544,7 @@
     {
       "FriendlyName": "Microsoft Edge",
       "AppId": "Microsoft.Edge",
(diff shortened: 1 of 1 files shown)


Raphire:
Heya, 

I think this is a good thing to point out. The message can probably just be simplified to something like:

```
Windows' default browser. WARNING: This also removed the only browser from Windows Sandbox.
```

As Win11Debloat can remove Edge even outside the EEA.

Raphire:
@AnonymousWP I tweaked the description a bit as noted in my earlier comment and merged it in :)

AnonymousWP:
All good, thanks. 🙂

PR #502, opened 2026-03-05
add options to disable Search Highlights and Photos AI features

CMD126:
Summary:

Added registry files to disable Windows Search Highlights and Microsoft Photos AI features.

Synced new feature flags with the script parameter system and default settings.

Details:

Added Disable_Search_Highlights.reg for current users and a Sysprep version to apply the IsDynamicSearchBoxEnabled=0 policy under the SearchSettings registry path.

Added Disable_Photo_AI_Features.reg (and Sysprep variant) to set DisableAIFeatures=1 under HKLM\SOFTWARE\Policies\Microsoft\Windows\Photos.

Updated Scripts/Get.ps1 with new switches: -DisablePhotoAI and -DisableSearchHighlights to match parameters in Win11Debloat.ps1.

Updated DefaultSettings.json to include DisableSearchHighlights: true as part of the default privacy configuration.

Files added:

Regfiles/Disable_Search_Highlights.reg

Regfiles/Sysprep/Disable_Search_Highlights.reg

Regfiles/Disable_Photo_AI_Features.reg

Regfiles/Sysprep/Disable_Photo_AI_Features.reg

Files updated:

Scripts/Get.ps1

DefaultSettings.json

### DIFF
--- a/Assets/Features.json
+++ b/Assets/Features.json
@@ -564,7 +564,7 @@
     },
     {
       "FeatureId": "DisableBing",
-      "Label": "Bing web search & Copilot integration",
+      "Label": "Bing web search & Copilot integration in search",
       "ToolTip": "This will turn off Bing web search results and Copilot integration in the Windows search experience. This feature uses policies, which will lock down certain settings.",
       "Category": "Start Menu",
       "Action": "Disable",
--- a/Schemas/MainWindow.xaml
+++ b/Schemas/MainWindow.xaml
@@ -1023,10 +1023,10 @@
                         <ColumnDefinition Width="*"/>
                         <ColumnDefinition Width="Auto"/>
                     </Grid.ColumnDefinitions>
-                    <Button x:Name="PreviousBtn" Grid.Column="0" Width="120" Height="36" Style="{DynamicResource SecondaryButtonStyle}" Visibility="Collapsed" Margin="10,0,0,0" AutomationProperties.Name="Previous">
+                    <Button x:Name="PreviousBtn" Grid.Column="0" Width="120" Height="36" Style="{DynamicResource SecondaryButtonStyle}" Visibility="Collapsed" Margin="10,0,0,0" AutomationProperties.Name="Back">
                         <StackPanel Orientation="Horizontal" VerticalAlignment="Center">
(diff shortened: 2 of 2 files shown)


Raphire:
Heya,

Do you have any sources/references for the registry keys / policies you used for disabling the AI options in Photos? I am unable to find them myself, setting the key also doesn't seem to remove the options:
image

As for the DisableSearchHighlights feature, this seems a bit redundant as this is already done when disabling bing/copilot from Windows Search.

CMD126:
Thanks for reviewing!

For the Photos AI policy, I found the key while inspecting registry activity with ProcMon when toggling AI features in Microsoft Photos. I’ll double-check whether it actually disables the UI or if the feature is controlled by another flag or server configuration. If it turns out the key isn’t effective, I’m happy to remove that part from the PR.

Regarding Search Highlights, that makes sense. I didn’t realize it was already implicitly disabled when Bing/Copilot search is turned off. I can drop that change as well to avoid redundancy.

CMD126:
You're right on both points, thanks for the review.

Photos AI: The registry key I used was speculative and doesn't actually work. I've removed it from the PR.

Search Highlights: I'd like to push back here. DisableSearchBoxSuggestions (Bing key) only blocks web results when you type. IsDynamicSearchBoxEnabled=0 is different, it controls the curated overlay (holiday themes, trending topics) that appears before you type anything. They are separate toggles in Settings > Privacy & security > Search permissions with different registry keys. Happy to drop it if you still feel it's redundant.

I also pushed two new additions:

DisableFindMyDevice: uses the documented Group Policy key (HKLM\SOFTWARE\Policies\Microsoft\FindMyDevice\AllowFindMyDevice=0). Kept separate from DisableLocationServices since some users want location for weather/maps but not Microsoft tracking their device position.

Lenovo and Dell OEM apps added to Apps.json (opt-in, SelectedByDefault: false): Lenovo Vantage, Lenovo Vantage Service, Dell SupportAssist, Dell Digital Delivery and Dell Mobile Connect. Only HP was covered before, this fills the gap for the two largest PC OEMs.

Raphire:
Heya,

I suspected as much, unfortunate as I would've liked to see that option.

As for the Search Highlights, I am not necessarily against a separate option. I suppose there is a use-case for those who don't want to outright disable web results.

I like the other additions, I see there's still registry keys missing for the Undo actions however.

PR #556, opened 2026-04-23
Added prompt asking user to run as adminstrator

David17c:
Made the script check if its running as adminsitrator if its not it asks the user if they want to allow it.


Raphire:
Heya,

Thanks for taking the time to contribute! I like this change, although I did make some slight alterations to ensure parameters are passed through when restarting as administrator.

PR #571, opened 2026-05-02
New Service Addition (Disc #547): Disable Update Service

tartaruslovesnoodles:
Hello, 

A little while back, I wrote [**"Improve Privacy and Update Control #547"**](https://github.com/Raphire/Win11Debloat/discussions/547) in your discussion board and suggested some changes. I've begun producing them, I decided to split these into 2 prs. This one covers the update service, while the other one will cover the privacy section. 

This one is a bit more polarizing and the most 'crude but pragmatic' solution of the bunch, My other solutions are much more easily streamlined by the system and already fit your scripts goals, so I think separating it from this PR is in all of our best interests. 

This all being said, I have tested it and it seems to work. However, I actually have windows 10, so my testing is not ironclad. I'd definitely recommend testing this before addition as I troubleshooted the process of creation on a Windows 10 machine.

### DIFF
--- a/Config/Features.json
+++ b/Config/Features.json
@@ -1359,6 +1359,21 @@
       "MinVersion": null,
       "MaxVersion": null
     },
+
+    {
+      "FeatureId": "Disablewuaserv",
+      "Label": "Disable Windows Update Service",
+      "ToolTip": "This will attempt to disable the updating service on a loop. When paired with other options like preventing idle detection, it can permanently stop updates.",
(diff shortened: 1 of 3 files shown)


Raphire:
Heya,

Thanks for taking the time to contribute. Unfortunately I do not think this is something I want to add to the script. Disabling the update service is something that I consider too risky and generally not worth it, outside of some niche cases.

PR #572, opened 2026-05-04
Update README.md

Solarauser123:
Simpler script to run


Raphire:
Heya,

While this is simpler to run, it does not support the command-line interface of the script. That's the reason the current command is the way it is. I'll think about this.

Raphire:
I am closing this for now, I think most people will be copying it anyway, but I will see if I can add this command to the documentation somewhere. Thanks!

PR #589, opened 2026-05-17
Prevent SSID Trilateration Through New Checkbox

tartaruslovesnoodles:
Hello, 

**Tl;dr** Inert powershell script Disables the wifi service when function called by delegator Win11Debloat.ps1, repeatedly checks if Wi-Fi is enabled and disables it if it is. Integrated with Functions.json, executechanges.ps1, [added file]DisableWifi.ps1, and Win11debloat.ps1

**Why beneficial:** 

I wrote [Discussion 547](https://github.com/Raphire/Win11Debloat/discussions/547) where I mentioned two potential features to add to your service. One of them already has a concluded PR, this is the other feature. 

The Wifi service is heavily logged by Microsoft, if it is enabled it will often negate turning off location services. You can actually test this through an interesting method: turn off location services on your host machine, then make a windows 10 virtual machine and leave 'find my location on' with your wifi still running on your host machine. In theory, it shouldn't be able to find your location, but it will, go on edge on the vm and look up "whats my location" and give the site permission. And yes to clarify this assumes the VM runs through its own adapter, ie the Virtualbox hypervisor's adapter.

Obviously I understand that not everyone can disable their Wifi, but this optional feature will be beneficial for those who can. 

**PS** If someone can't use ethernet the user could also bypass Wi-fi through a usb tether with a secure device (ie a phone) connected to the wi-fi the computer would've connected to. A "pseudo-wired" connection.

### DIFF
--- a/Config/Features.json
+++ b/Config/Features.json
@@ -374,6 +374,20 @@
       "MinVersion": null,
       "MaxVersion": null
     },
+
+    {
+      "FeatureId": "DisableWiFi",
+      "Label": "Disable Wi-Fi Service (prevent location leaking when disabled).",
+      "ToolTip": "Even with location services disabled and going through extensive effort, it is all in vain if you leave Wi-Fi and Wi-Fi discovery enabled. Wi-Fi trilateration via discoverable SSIDs nearby can leak your location.",
+      "Category": "Privacy & Suggested Content",
+      "RegistryKey": null,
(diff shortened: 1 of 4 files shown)


Raphire:
Heya,

Thanks for taking the time to contribute to Win11Debloat. I'm not sure if this feature is suitable for the script, as it seems quite niche. I also don't think this implementation will work as it would require the script to run all the time, while the rest of the script is more intended to be strictly run once.

PR #609, opened 2026-06-07
Add Support for "-user" Parameter running under SYSTEM

soccerzockt:
Handle offline and currently active Users, when using the -user Parameter, when running with the same or different account than the currently logged in.


Raphire:
@soccerzockt Thanks for taking the time to create the PR. I really like the idea, it was well over-due to implement this.

I noticed that a few changes were reverted, probably because you based your changes on an older version, so I fixed that real quick. 

I have also refactored how the registry changes were applied to the loaded hive, the import function already had the capability to parse and directly write powershell changes. I co-opted that to write the changes to the loaded hive. (75ce9ae)

soccerzockt:
@Raphire Thank you for taking such swift action on feedback and recommendations! Generally great work!

Sorry for the PR, not familiar with git (yet).
Looking forward to trying the new release as soon as ist released. :)

Raphire:
> @Raphire Thank you for taking such swift action on feedback and recommendations! Generally great work!
> 
> Sorry for the PR, not familiar with git (yet). Looking forward to trying the new release as soon as ist released. :)

No worries, it was easy to sort out! I hope to have the new release ready some time next week.

PR #613, opened 2026-06-08
feat(registry): implement automatic registry rollback on execution failure

### LINKED ISSUE #612: Enhancement: Implement automatic registry rollback on execution failure
Currently, if the script fails or encounters an unhandled exception during the execution of debloat parameters, the registry is left in a partially modified, inconsistent state.

This enhancement wraps the execution loop in a try-catch block. If an execution failure occurs, the script automatically loads the pre-execution registry backup (created via New-RegistrySettingsBackup) and restores the registry to its original state, ensuring transactional integrity.

HetCreep:
Resolves #612

### Changes
- Intercepts script execution failures during parameter execution.
- If a failure is caught, automatically restores the pre-execution registry state using the JSON backup file via Restore-RegistryBackupState.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved recovery when applying debloat changes fails by capturing the pre-execution registry backup path from `New-RegistrySettingsBackup`, wrapping the “execute actionable parameters” and “execute undo operations” loops in a `try`/`catch`, and triggering an automatic rollback via `Restore-RegistryBackupState` when parameter execution fails or when `$script:RegistryImportFailures` increases during either phase; rollback errors are logged and the original failure is re-thrown.

### DIFF
--- a/Scripts/Features/Invoke-Changes.ps1
+++ b/Scripts/Features/Invoke-Changes.ps1
@@ -405,8 +405,12 @@ function Invoke-AllChanges {
     }
 
     # ================================================================
-    # Phase 3 & 4: Apply and undo features, with automatic registry
-    # rollback to the pre-execution backup if a registry import fails.
+    # Phase 3 & 4: Apply and undo features. On a registry import failure,
+    # point the user at the pre-execution backup rather than reverting it
+    # automatically -- an unattended rollback can hit the same access
+    # failure that broke the apply in the first place, and reverting
+    # registry state without asking is not a call the script should make
+    # for the user.
     # ================================================================
     $initialFailures = $script:RegistryImportFailures
     try {
@@ -434,15 +438,8 @@ function Invoke-AllChanges {
         Write-Error "Execution failed: $_"
 
         if ($backupFile -and (Test-Path -LiteralPath $backupFile)) {
-            Write-Warning "Attempting automatic registry rollback using backup: $backupFile"
-            try {
-                $backupData = Import-RegistryBackup -FilePath $backupFile
-                Restore-RegistryBackupState -Backup $backupData
-                Write-Host "Registry successfully rolled back to pre-execution state." -ForegroundColor Green
-            }
-            catch {
-                Write-Error "Registry rollback failed: $_"
-            }
+            Write-Warning "A pre-execution registry backup was saved to: $backupFile"
+            Write-Warning "Use the 'Restore Backup' option in the app to revert the changes made in this run."
         }
 
         throw
--- a/Scripts/Features/Invoke-Changes.ps1
+++ b/Scripts/Features/Invoke-Changes.ps1
@@ -352,6 +352,7 @@ function Invoke-AllChanges {
     if ($needsBackup -and -not $script:Params.ContainsKey('SkipRegistryBackup')) { $totalSteps++ }
     if ($script:Params.ContainsKey("CreateRestorePoint")) { $totalSteps++ }
     $step = 0
+    $backupFile = $null
 
     # ================================================================
     # Phase 1: Registry backup
@@ -375,7 +376,7 @@ function Invoke-AllChanges {
                         [PSCustomObject]@{ FeatureId = $_; RegistryKey = (Resolve-UndoRegFilePath $f.RegistryUndoKey) }
                     }
                 } | Where-Object { $_ })
-                New-RegistrySettingsBackup -ActionableKeys $applyIds -ExtraFeatures $undoSyntheticFeatures | Out-Null
+                $backupFile = New-RegistrySettingsBackup -ActionableKeys $applyIds -ExtraFeatures $undoSyntheticFeatures
             }
             catch {
                 throw "Registry backup failed before applying changes. $($_.Exception.Message)"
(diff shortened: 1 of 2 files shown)


Raphire:
Heya,

I do like this idea, I'll take a look at this soon when I have time. Thanks!

Raphire:
@coderabbitai review

HetCreep:
Heads-up on the merge conflict here: #641 refactored `ExecuteChanges.ps1` into `InvokeChanges.ps1` and split the execute/undo loops out into `Invoke-ApplyFeatures` / `Invoke-UndoFeatures` (called from `Invoke-AllChanges`). This PR's rollback wraps those loops in a single try/catch inside the old monolithic function, so it needs re-applying onto the new phased structure rather than a mechanical rebase.

Since it's the registry-rollback safety path and I can't run-test the restore on my end, I'd rather not force-push an untested rewrite of it. Would you prefer I redo it against the new structure (you'd want to verify the rollback/restore path before merge), or would you rather fold it into your own refactor? Either works for me -- just didn't want to decide that for you.

Raphire:
> Heads-up on the merge conflict here: #641 refactored `ExecuteChanges.ps1` into `InvokeChanges.ps1` and split the execute/undo loops out into `Invoke-ApplyFeatures` / `Invoke-UndoFeatures` (called from `Invoke-AllChanges`). This PR's rollback wraps those loops in a single try/catch inside the old monolithic function, so it needs re-applying onto the new phased structure rather than a mechanical rebase.
> 
> Since it's the registry-rollback safety path and I can't run-test the restore on my end, I'd rather not force-push an untested rewrite of it. Would you prefer I redo it against the new structure (you'd want to verify the rollback/restore path before merge), or would you rather fold it into your own refactor? Either works for me -- just didn't want to decide that for you.

Heya,

That's a fair assessment. If you'd be open to redoing it that would be great. I do apologize for the issues caused by the recent refactoring going on with the script, it was something that was sorely needed to keep things clear and well structured. As this feature is quite impactful I wasn't satisfied yet to merge it before fully testing.

HetCreep:
Redone against the new structure, as discussed. The rollback now lives in `Invoke-AllChanges`: it captures the path returned by `New-RegistrySettingsBackup`, wraps the apply (`Invoke-ApplyFeatures`) and undo (`Invoke-UndoFeatures`) phases in a try/catch, throws when `RegistryImportFailures` increases, and on failure restores the pre-execution state via `Load-RegistryBackupFromFile` + `Restore-RegistryBackupState` before rethrowing. WhatIf is unaffected -- no backup is created in WhatIf mode, so no rollback runs.

It's parse-clean, but I haven't run-tested the restore path on my end (no debloat run here), so please verify the rollback/restore behaviour before merge, as you mentioned. (Apologies for the brief close/reopen above -- a force-push hiccup on my side while resetting the branch onto the new structure.)

STiFLeR7:
Reviewed this against the registry restore/import code paths it touches (I was just working on #686 / #704 in the restore path, which is directly relevant here). Two grounded findings, one blocking-ish and one informational:

**1. The failure-detection trigger has a silent blind spot for partial access-denial.**

The rollback only fires when `$script:RegistryImportFailures -gt $initialFailures` (lines added around `Invoke-ApplyFeatures`/`Invoke-UndoFeatures`). That counter is only incremented in `ImportRegistryFile`'s outer `catch` (`Scripts/Features/ImportRegistryFile.ps1:109-113`), which only fires if the import actually throws. But `Invoke-RegistryOperationsFromRegFile` (`Scripts/Helpers/ApplyRegistryRegFile.ps1:196-228`) only throws when **all** operations in a `.reg` file are access-denied (`$accessDeniedCount -eq $totalOperations`) — on a *partial* denial it just `Write-Warning`s and returns normally, no exception. So on a GPO-hardened/MDM-managed machine where some keys are writable and others aren't, some registry changes can genuinely fail to apply without `$script:RegistryImportFailures` ever incrementing — meaning this PR's whole point (auto-rollback on failure) silently won't trigger for that failure mode. Might be worth having `Invoke-RegistryOperationsFromRegFile` return (or the caller inspect) the access-denied count directly rather than relying only on the throw/no-throw signal.

**2. The rollback mechanism itself reuses a currently-unsafe restore path.**

`Restore-RegistryBackupState -Backup $backupData` (called in the new `catch` block) delegates to `Restore-RegistryKeySnapshot`, which deletes the live registry subtree before rewriting it, with no rollback of its own if the rewrite throws partway (see #686). I just opened #704, which closes the "corrupted/hand-edited backup" trigger for that bug (Data/Kind validation before any restore begins) — but the backup this PR generates is a fresh, in-memory snapshot of pre-execution state, not user-edited, so #704 is unlikely to matter for *this* call site specifically. The remaining trigger from #686 (a runtime `SetValue` failure mid-restore, e.g. permission drift between backup and rollback) is still a real, if narrow, risk for this PR's rollback path — worth being aware that a mid-rollback failure here has the same "already deleted, not yet rewritten" exposure as a mid-restore failure in the manual Restore Backup flow.

Neither of these is necessarily a reason to block — #613's overall direction (backup before apply, roll back on detected failure) is a real safety improvement over the status quo of no rollback at all — but I wanted to flag both gaps explicitly since they're easy to miss without having just traced through the restore path.

HetCreep:
Rebased onto current master: re-applied the rollback onto the renamed Invoke-Changes.ps1 (post #708), switched the restore call to the renamed Import-RegistryBackup, and preserved the #700 cancellation check between the apply and undo phases (a cancel returns without triggering rollback, since cancellation is not a failure). Parse-clean; I could not run the new Pester suite locally (in-box Pester 3.4, the suite needs v5 and I avoid installing tools on this machine) -- the tests.yml workflow should exercise it on this PR.

On STiFLeR7's two findings above -- both are accurate, and worth stating how they relate to this PR's scope: (1) the partial-access-denial blind spot lives in Invoke-RegistryOperationsFromRegFile's throw-only-on-total-denial contract, so the right fix is surfacing the denied count from that helper (which would then flow through this PR's failure detection for free) rather than widening this PR's trigger heuristically; (2) the rollback path inheriting #686's delete-before-rewrite exposure is real -- with #704 merged, the corrupted-backup trigger is now validated out, and since the backup here is freshly tool-generated in the same run, the residual risk is limited to a runtime SetValue failure mid-rollback (narrow, but real until #686's case 2 gets a mitigation design). Both feel like follow-ups at the restore-layer level rather than blockers for adding rollback-on-failure at all, but that's your call.

Raphire:
Thanks @STiFLeR7 and @HetCreep! I think the second issue should be fixed now as of #710, the script should now handle those restores much more conservatively and with some extra validation, it no longer deletes keys that could potentially not get restored if something fails along the way.

I will fully review this when I have more time this week.

Raphire:
This generally looks fine from a technical perspective, although I am worried a rollback to the backup might fail for the same reason the original apply action failed. Looking at it from that perspective this might not really add much.

I also don't think we should automatically rollback any changes without prompting the user, to let them choose.

What are your thoughts on that?

HetCreep:
Heya,

Both points land -- pivoted this to match.

**On the rollback-reliability point:** confirmed by reading through the restore path (`Restore-RegistryKeySnapshotAtPath` in `Restore-RegistryApplyState.ps1`, post #710) -- for whichever key failed to apply, the revert would call the identical `SetValue`/`CreateSubKey` against the identical ACL, so it hits the identical wall. An unattended rollback isn't a reliable safety net for the failure mode it exists to catch.

(Aside, not blocking this PR: while tracing that path I noticed `Restore-RegistryBackupState`'s restore loop has no per-key isolation -- one exception on any root snapshot aborts the remaining ones too, so a manual Restore Backup that hits a bad key partway through would silently stop reverting the rest. Separate from this PR since it's the existing GUI restore path, happy to open an issue for it if useful.)

**On the "don't auto-revert without asking" point:** agreed, and it also sidesteps the reliability problem above entirely -- no rollback is attempted, so there's nothing to fail. On a registry import failure the script now reports the backup file location and points to the existing Restore Backup option instead of touching the registry again on its own.

Pushed (`17abdf5`). Updated the two rollback specs to match -- parse-clean, but same as last round I couldn't run the v5 Pester suite locally; tests.yml should cover it.

PR #617, opened 2026-06-08
fix(appx): expose swallowed exceptions during Appx Package uninstallation via Write-Verbose

### LINKED ISSUE #616: Improve Observability: Expose swallowed exceptions during Appx Package uninstallation
In the Appx package uninstallation pipeline, exceptions thrown during package removal are swallowed in the try-catch block. If the removal fails, the script outputs a generic "Something went wrong while trying to remove <app>" message, making it difficult to diagnose why the removal failed (e.g. access denied, package not found, or dependency blocks).

This enhancement logs the exception message via Write-Verbose in the catch block to improve observability and debugging.

HetCreep:
Resolves #616

### Changes
- Adds a Write-Verbose call inside the catch block of RemoveApps to output the exact exception details.

### DIFF
--- a/Scripts/AppRemoval/RemoveApps.ps1
+++ b/Scripts/AppRemoval/RemoveApps.ps1
@@ -124,6 +124,7 @@ function RemoveApps {
             }
         }
         catch {
+            Write-Verbose "Something went wrong while trying to remove $app: $_"
             if ($DebugPreference -ne "SilentlyContinue") {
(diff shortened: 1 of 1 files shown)


Raphire:
Heya. The existing logging already did output the exception, although in a round about way. I have cleaned up the old logging and fixed an issue with the app name variable in the logging string.

PR #625, opened 2026-06-14
Fixed the system restore point creation logic

aadvt:
## Summary

While using Win11Debloat on my own system, I selected the option to create a restore point before making changes. Later, when I wanted to roll back, I found that no restore point was available.

After looking into the restore point creation logic, I found a few issues that could prevent restore points from being created correctly:

* `RPSessionInterval` was being used to determine whether System Restore was enabled, even though it only controls restore point throttling.
* Missing registry values could cause errors.
* Systems with no existing restore points could fail before creating their first restore point.

## Changes

* Replaced the `RPSessionInterval` check with a `DisableSR` check.
* Added handling for missing registry values.
* Added handling for systems with no existing restore points.

## Testing

Tested locally on Windows 11.

* Removed existing restore points.
* Loaded and ran the modified script.
* Verified that a new restore point was created successfully.

Everything worked as expected.

### DIFF
--- a/Scripts/Features/CreateSystemRestorePoint.ps1
+++ b/Scripts/Features/CreateSystemRestorePoint.ps1
@@ -1,8 +1,10 @@
 function CreateSystemRestorePoint {
-    $SysRestore = Get-ItemProperty -Path "HKLM:\SOFTWARE\Microsoft\Windows NT\CurrentVersion\SystemRestore" -Name "RPSessionInterval"
+    # Check if System Restore is disabled by reading the DisableSR registry value.
+    # DisableSR = 1 means disabled. If the value is missing or 0, System Restore is enabled.
+    $disableSR = (Get-ItemProperty -Path "HKLM:\SOFTWARE\Microsoft\Windows NT\CurrentVersion\SystemRestore" -Name "DisableSR" -ErrorAction SilentlyContinue).DisableSR
     $failed = $false
 
-    if ($SysRestore.RPSessionInterval -eq 0) {
+    if ($disableSR -eq 1) {
         # In GUI mode, skip the prompt and just try to enable it
         if ($script:GuiWindow -or $Silent -or $( Read-Host -Prompt "System restore is disabled, would you like to enable it and create a restore point? (y/n)") -eq 'y') {
             try {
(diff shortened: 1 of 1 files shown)


Raphire:
Heya,

Unfortunately `DisableSR` doesn't seem to be the right value to check if System Restore point is enabled. It ignores cases where system restore point as a feature is enabled globally, but disabled on all drives. In my testing on your PR it does not detect that System Restore is not enabled on the drive, and it fails to create a restore point even though it seemed to succeed.

RPSessionInterval seems to be the only key that consistently works in my testing, and according to many others on the internet. See [here](https://www.ninjaone.com/script-hub/get-system-restore-status/), and [here](https://www.reddit.com/r/pdq/comments/15gcrtj/checking_for_system_restore_being_enabled/) for a few examples.

The changes to how $recentRestorePoints is determined seem unnecessary as it already handles cases of 0 restore points just fine in my testing.

aadvt:
Hey, thanks for taking the time to review and test it.

That makes sense. My assumption was that `DisableSR` would be a more reliable indicator than `RPSessionInterval`, but I hadn't considered the case where System Restore is enabled globally while protection is disabled on the drive itself.

Regarding the `Get-ComputerRestorePoint` changes, I was able to reproduce a situation on my machine where no restore point was available after running the script, which is what led me down this path. If you're unable to reproduce that behavior, then I may have been looking at a different underlying issue.

Out of curiosity, would checking whether System Protection is enabled on the target drive be a better direction to investigate here? I'm still interested in figuring out what caused the restore point creation to fail on my system originally.

Thanks again for the feedback!!! :)

Raphire:
> Regarding the Get-ComputerRestorePoint changes, I was able to reproduce a situation on my machine where no restore point was available after running the script, which is what led me down this path. If you're unable to reproduce that behavior, then I may have been looking at a different underlying issue.

Do you have a log to share that indicates it threw an error there for you? I personally have not encountered that, but I'm certainly interested to investigate it.

> Out of curiosity, would checking whether System Protection is enabled on the target drive be a better direction to investigate here? I'm still interested in figuring out what caused the restore point creation to fail on my system originally.

I think this would be a better direction to look yes. I would be interested to see your logs, maybe they show some more details on the issue.

aadvt:
Okay so it looks like my original assumption was wrong. I initially started looking into this because I had used the restore point option before running Win11Debloat a while back, and when I later wanted to roll back while troubleshooting a freezing issue, I couldn't find any restore point available. That led me down the path of investigating the creation logic.

At this point, I can confirm that restore point creation is working in my testing. I'm still not sure what happened to the restore point I expected to have from my earlier use of the tool, but it doesn't seem to be caused by the issue I was trying to address in this PR.

Thanks for taking the time to review this. I learned quite a bit digging into it further.

Raphire:
> Okay so it looks like my original assumption was wrong. I initially started looking into this because I had used the restore point option before running Win11Debloat a while back, and when I later wanted to roll back while troubleshooting a freezing issue, I couldn't find any restore point available. That led me down the path of investigating the creation logic.
> 
> At this point, I can confirm that restore point creation is working in my testing. I'm still not sure what happened to the restore point I expected to have from my earlier use of the tool, but it doesn't seem to be caused by the issue I was trying to address in this PR.
> 
> Thanks for taking the time to review this. I learned quite a bit digging into it further.

Windows does sometimes delete existing restore points after an update, it's possible that this happened in your case.

aadvt:
That would make sense. Looking back, the freezing issues did start shortly after a Windows update, so it's definitely possible the restore point was removed as part of that process.

In any case, thanks again for taking the time to review and test this. 

I'll go ahead and close the PR since the issue I was trying to address doesn't appear to be the root cause.

PR #628, opened 2026-06-15
fix(appx): pre-fetch installed packages to prevent redundant uninstallation logs

### LINKED ISSUE #627: Improvement: Optimize Appx uninstallation console output and prevent redundant logging
Currently, when executing the script's Appx removal features, the console prints `Removing <AppId>` for every app in the removal list, regardless of whether the app is actually installed on the system or has already been uninstalled. This clutters the console logs with redundant and misleading logs.

This improvement:
1. **Pre-fetch Appx cache**: Pre-fetches the list of currently installed and provisioned packages in a single fast batch query at the beginning of the removal function.
2. **Existence checks**: Checks each app against the pre-fetched cache, skipping the uninstallation commands and avoiding printing `Removing <AppId>` if the package is already missing.
3. **Query Success Fallback**: Gracefully degrades by falling back to standard uninstallation if the pre-fetch query fails (e.g. registry/service errors), preventing silent uninstallation skips.

HetCreep:
Resolves #627

### Changes
- Pre-fetches installed and provisioned Appx packages in a single batch to optimize performance.
- Checks if the target app package is present in the cache before logging or initiating uninstallation.
- Skips already-uninstalled packages to keep console logs clean and accurate.
- Gracefully degrades by executing normal uninstallation steps if the pre-fetch query fails.

## Summary

The `RemoveApps` function now pre-fetches the complete set of installed and (for `AllUsers`) provisioned Appx packages at the start of the removal process using a non-blocking `Invoke-NonBlocking` call. The pre-fetch is wrapped in try/catch and stores results in `$installedApps`, `$provisionedApps`, along with a `$querySuccess` flag.

During the per-app loop, removal is gated by this cache. When `$querySuccess` is true, each requested app is checked against the pre-fetched lists to determine whether it is already uninstalled—skipping the WinGet/`Remove-AppxPackage` logic (and its associated console “Removing <AppId>” output) for apps that are not present. For `AllUsers`, the script additionally consults `$provisionedApps` when the app is not found in installed packages. Presence detection includes special-casing for `Microsoft.OneDrive` and Microsoft Edge identifiers to preserve existing removal behavior for those apps.

Graceful degradation is included: if the pre-fetch query fails due to registry/service errors (`$querySuccess` remains false), the script falls back to the existing uninstallation flow rather than silently skipping packages.

**Lines changed:** +56/-0

### DIFF
--- a/Scripts/AppRemoval/RemoveApps.ps1
+++ b/Scripts/AppRemoval/RemoveApps.ps1
@@ -37,6 +37,44 @@ function RemoveApps {
     # Determine target from script-level params, defaulting to AllUsers
     $targetUser = GetTargetUserForAppRemoval
 
+    # Pre-fetch installed and provisioned packages to optimize performance
+    $installedApps = @()
+    $provisionedApps = @()
+    $querySuccess = $false
+
+    $packagesInfo = Invoke-NonBlocking -ScriptBlock {
+        param($target, $user)
+        $inst = @()
+        $prov = @()
+        $success = $false
+        try {
+            switch ($target) {
+                "AllUsers" {
+                    $inst = @(Get-AppxPackage -AllUsers -ErrorAction Stop | Select-Object -ExpandProperty Name)
+                    $prov = @(Get-AppxProvisionedPackage -Online -ErrorAction Stop | Select-Object -ExpandProperty PackageName)
+                }
+                "CurrentUser" {
+                    $inst = @(Get-AppxPackage -ErrorAction Stop | Select-Object -ExpandProperty Name)
+                }
+                default {
+                    $userAccount = New-Object System.Security.Principal.NTAccount($user)
+                    $userSid = $userAccount.Translate([System.Security.Principal.SecurityIdentifier]).Value
+                    $inst = @(Get-AppxPackage -User $userSid -ErrorAction Stop | Select-Object -ExpandProperty Name)
+                }
+            }
+            $success = $true
+        } catch {
+            Write-Verbose "Appx package pre-fetch query failed: $_"
+        }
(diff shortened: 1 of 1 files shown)


HetCreep:
Heads-up after the #662 rewrite of `RemoveApps.ps1`: the specific clutter #627 describes is still present -- `Write-Host "Removing $app"` (RemoveApps.ps1:60) fires for every app in the list before any installed-check, so the `Removing <AppId>` lines still print for apps that aren't installed. So #627 itself is still valid.

This PR's pre-fetch implementation, however, no longer applies -- it was written against the old monolithic `RemoveApps`, which #662 fully restructured (`Remove-AppxApp` / `Get-AppRemovalMethod` / per-app removal). I'm happy to reimplement the existence-check-before-logging on the new structure once the app-removal refactor settles (post #662 / #666), or feel free to close this in the meantime -- whichever you prefer.

HetCreep:
Closing this -- it's superseded by the `RemoveApps` rewrite in #662. This PR's pre-fetch was written against the old monolithic `RemoveApps`, which #662 split into separate functions (`Remove-WinGetApp` / `Remove-AppxApp` / `Get-AppRemovalMethod` / ...) and which already does its own installed-state checking (`Test-AppStillInstalled` + the `GetInstalledAppsViaWinget` pre-fetch). The original concern in #627 (the unconditional `Removing <AppId>` line, still at `RemoveApps.ps1:60`) remains valid against the new code, but it would need reimplementing on the new structure rather than this diff. Thanks!

Raphire:
Heya,

While I do like this idea functionally, I do think it should probably hook into the other functionality that checks against installed apps. It would preferably be done once, maybe in the background when the app starts. This could then be used globally, in the UI, and in the app removal logic as was touched with this PR. This would require a new approach and would introduce an additional layer of complexity.

PR #649, opened 2026-06-21
fix(app-removal): block removal of shell/sign-in/runtime-critical packages

### LINKED ISSUE #648: Custom app-removal list can remove shell-critical packages (no guardrail / brick risk)
### Summary
The app-removal path has no guardrail against removing packages critical to the Windows shell, sign-in, or app runtime. A bad entry can leave the machine broken (no Start menu / taskbar, broken sign-in, apps that won't launch).

### How it can happen
`RemoveApps` wraps each entry as a substring wildcard and removes it for all users:
```powershell
$appPattern = '*' + $app + '*'
Get-AppxPackage -Name $appPattern -AllUsers | Remove-AppxPackage -AllUsers
```
The `RemoveAppsCustom` path reads `Config\CustomAppsList` (`LoadAppsFromFile`) and passes it straight to `RemoveApps` **without** `ValidateAppslist` (only the `-Apps` path validates). So:
- a hand-edited / externally-written `CustomAppsList` containing e.g. `Microsoft.Windows.ShellExperienceHost`, `Microsoft.AAD.BrokerPlugin`, or `Microsoft.VCLibs`, or
- an over-broad entry like `Microsoft.Windows` (substring-wildcarded to `*Microsoft.Windows*`)

reaches `Remove-AppxPackage -AllUsers` with no safety net.

### Fix
A hard denylist in `RemoveApps` that skips shell/sign-in/runtime-critical packages regardless of how they entered the list (covers every removal path). PR linked below. The denylisted families (ShellExperienceHost, AAD.BrokerPlugin, Microsoft.VCLibs, Microsoft.WindowsAppRuntime, DesktopAppInstaller, .) do not appear in `Config/Apps.json`, so legitimate removals are unaffected.

> Found via static analysis; not yet run on Windows.

HetCreep:
Fixes #648

Adds a hard denylist in `RemoveApps` so packages critical to the shell, sign-in, or app runtime are never removed - regardless of how they entered the list. The `-Apps` path validates against `Apps.json`, but `RemoveAppsCustom` reads `CustomAppsList` unchecked and entries are substring-wildcarded, so this is defense-in-depth against bricking the machine.

The denylisted families (`ShellExperienceHost`, `AAD.BrokerPlugin`, `Microsoft.VCLibs`, `Microsoft.WindowsAppRuntime`, `DesktopAppInstaller`, .) do not appear in `Config/Apps.json`, so no legitimate removal is affected.

Static check:
```
'Microsoft.Windows.ShellExperienceHost' -> blocked
'Microsoft.AAD.BrokerPlugin'            -> blocked
'Microsoft.VCLibs.140.00'               -> blocked
'Microsoft.WindowsStore'                -> allowed
'Microsoft.BingNews'                    -> allowed
```
File parses clean.

> ?? Static analysis only - not run on Windows. Runtime confirmation welcome before merge.

## RemoveApps.ps1

Added a `$criticalAppDenylist` hard safety guard in `RemoveApps` to block removal of Windows system-essential app packages. During iteration over `$appsList`, each `$app` is checked against the denylist using a substring wildcard match (`$app -like "*$criticalApp*"`). On a match, `RemoveApps` logs a yellow “Skipping …” message and immediately `continue`s, preventing any existing OneDrive/Edge WinGet (and scheduled-task) uninstall logic and any `Remove-AppxPackage` / `Remove-ProvisionedAppxPackage` removal work from running.

The denylist contains these 12 critical package name fragments:

- `ShellExperienceHost`, `StartMenuExperienceHost`
- `AAD.BrokerPlugin`, `CloudExperienceHost`
- `Microsoft.UI.Xaml`
- `Microsoft.VCLibs`, `Microsoft.NET.Native`, `Microsoft.WindowsAppRuntime`
- `DesktopAppInstaller`
- `Microsoft.AccountsControl`, `Microsoft.LockApp`
- `SecHealthUI`

This protection applies to both `RemoveApps` and the custom-list flow (`RemoveAppsCustom`) because `RemoveAppsCustom` loads entries from `CustomAppsListFilePath` and passes them directly to `RemoveApps`, which performs the same denylist filtering for every entry.

Lines changed: +17/-0

### DIFF
--- a/Scripts/AppRemoval/RemoveApps.ps1
+++ b/Scripts/AppRemoval/RemoveApps.ps1
@@ -43,12 +43,29 @@ function RemoveApps {
     $edgeUninstallSucceeded = $false
     $edgeScheduledTaskAdded = $false
 
+    # Packages that must never be removed: removing them breaks the shell, sign-in, or the app runtime
+    $criticalAppDenylist = @(
+        'ShellExperienceHost', 'StartMenuExperienceHost', 'AAD.BrokerPlugin', 'CloudExperienceHost',
+        'Microsoft.UI.Xaml', 'Microsoft.VCLibs', 'Microsoft.NET.Native', 'Microsoft.WindowsAppRuntime',
+        'DesktopAppInstaller', 'Microsoft.AccountsControl', 'Microsoft.LockApp', 'SecHealthUI'
+    )
+
     Foreach ($app in $appsList) {
         if ($script:CancelRequested) {
             return
         }
 
         $appIndex++
+
+        # Safety guard: never remove packages critical to the shell, sign-in, or app runtimes
+        $isCriticalApp = $false
+        foreach ($criticalApp in $criticalAppDenylist) {
+            if ($app -like "*$criticalApp*") { $isCriticalApp = $true; break }
+        }
+        if ($isCriticalApp) {
+            Write-Host "Skipping '$app': this package is critical to Windows and removal is blocked." -ForegroundColor Yellow
+            continue
+        }
 
         # Update step name and sub-progress to show which app is being removed (only for bulk removal)
         if ($script:ApplySubStepCallback -and $appCount -gt 1) {


Raphire:
Heya,

I'm not sure if this is really necessary. The CustomAppsList is a legacy format that I will likely deprecate with the next release. Outside of that all app removal should be properly checked against allowed appids from `Apps.json`. The substring wildcard is also something i want to remove in the future as it could indeed result in unintended removal of apps if they happen to match.

I'd rather focus on that then have to maintain a list of apps that guards against a user adding unsupported apps in a file that is not meant to be edited by the user.

HetCreep:
That makes sense -- dropping the `*$app*` wildcard and deprecating the CustomAppsList fixes the root cause directly, which the denylist was only ever guarding against. I'll defer to your approach -- feel free to close this, or keep it as an interim guard until that lands, whichever you prefer.

Raphire:
CustomAppsList support has been removed now, and appid handling has been tightened.

PR #653, opened 2026-06-21
fix(start-menu): don't overwrite the original backup on repeated runs

### LINKED ISSUE #652: Running Clear/Replace Start menu twice destroys the original layout backup
### Summary
Running **Clear Start menu / Replace Start menu twice permanently destroys the user's original Start layout backup**, so a later restore brings back the cleared/template layout instead of the original.

### Root cause
`ReplaceStartMenu` (`Scripts/Features/ReplaceStartMenu.ps1`) backs up the current `start2.bin` on every run:
```powershell
$backupBinFile = $startMenuBinFile + ".bak"
if (Test-Path $startMenuBinFile) {
    Move-Item -Path $startMenuBinFile -Destination $backupBinFile -Force
}
```
- **Run 1:** `start2.bin` (the user's original) ? moved to `start2.bin.bak`. Correct.
- **Run 2:** `start2.bin` is now the **replaced** (cleared/template) layout. `Move-Item -Force` overwrites `start2.bin.bak` with it ? the original is gone.

`RestoreStartMenuFromBackup` then restores the cleared layout. Data loss is permanent.

### Fix
Only create the backup when one doesn't already exist; keep the original from run 1. PR linked.

> Found via static analysis on Windows 11.

HetCreep:
Fixes #652

`ReplaceStartMenu` backed up the current `start2.bin` to `start2.bin.bak` with `Move-Item -Force` on **every** run. On the second run the current `start2.bin` is the already-replaced layout, so `-Force` overwrote the `.bak` that held the user's **original** layout - losing it permanently and making `RestoreStartMenuFromBackup` restore the cleared layout.

Now the backup is created only when one doesn't already exist; otherwise the current (replaced) file is removed and the original `.bak` is preserved. File parses clean.

> ?? Static analysis only - not run on Windows. Runtime confirmation welcome.

This fix addresses a critical data loss bug in the start menu backup workflow. `ReplaceStartMenu` previously backed up `start2.bin` by moving it to `start2.bin.bak` with `-Force` on every run, allowing a second execution to overwrite the original backup with the already-replaced (cleared/template) layout—so `RestoreStartMenuFromBackup` would restore the wrong data.

The updated `ReplaceStartMenu` preserves the first backup:
- If `start2.bin` exists and `start2.bin.bak` does not, it moves `start2.bin` to `start2.bin.bak` (first run).
- If `start2.bin.bak` already exists, it deletes the current `start2.bin` instead of overwriting the existing backup.

After ensuring the correct backup behavior, the function continues by copying the template `start2.bin` into place as before, preventing repeated runs from permanently destroying the user’s original start menu layout.

### DIFF
--- a/Scripts/Features/ReplaceStartMenu.ps1
+++ b/Scripts/Features/ReplaceStartMenu.ps1
@@ -74,8 +74,16 @@ function ReplaceStartMenu {
     $backupBinFile = $startMenuBinFile + ".bak"
 
     if (Test-Path $startMenuBinFile) {
-        # Backup current start menu file
-        Move-Item -Path $startMenuBinFile -Destination $backupBinFile -Force
+        if (Test-Path $backupBinFile) {
+            # A backup already exists from a previous run, so the current start2.bin is the
+            # already-replaced layout, not the user's original. Don't overwrite the original
+            # backup - just remove the current file (the template is copied in below).
+            Remove-Item -Path $startMenuBinFile -Force
+        }
+        else {
+            # Backup the original start menu file (first run only)
+            Move-Item -Path $startMenuBinFile -Destination $backupBinFile -Force
+        }
     }
     else {
         Write-Host "Unable to find original start2.bin file for user $userName, no backup was created for this user" -ForegroundColor Yellow


Raphire:
Heya,

I think this approach creates new problems, someone may run the script for a second time months later and not get their start menu backed up.

I think this should probably get the same treatment as registry backups: Create a backup with the date/time embedded in the name (Example: `Win11Debloat-Start2BinBackup-YYYYMMDD_HHMMss.bak`). This would require a rework of the start menu restore function, as this relies on the hardcoded backup location, but I think we can just remove the auto discover backup feature and let the user always select the backup.

HetCreep:
Good point -- not overwriting the backup just trades one problem for another (a later run gets no fresh backup). The timestamped approach matching the registry backups is the better fix; I'll defer to your rework. Happy to help with the dated-backup + restore changes if useful -- otherwise feel free to close.

Raphire:
I have opened #666 to implement it.

PR #655, opened 2026-06-21
fix(threading): surface runspace errors instead of swallowing them in GUI mode

### LINKED ISSUE #654: GUI-mode operations fail silently: Invoke-NonBlocking swallows all runspace errors
### Summary
In **GUI mode**, operations that fail with non-terminating errors fail **silently** - no message, no log, nothing in the UI. Affects app removal, winget uninstall, scheduled-task disable, and optional-feature toggles.

### Root cause
`Invoke-NonBlocking` runs the scriptblock in a separate `[powershell]` runspace in GUI mode and calls `$ps.EndInvoke($handle)`, but **never inspects `$ps.HadErrors` / `$ps.Streams.Error`**. A runspace keeps its own error stream; it is discarded on `$ps.Dispose()`. Callers use `-ErrorAction Continue` (e.g. `RemoveApps.ps1`: `Remove-AppxPackage -AllUsers -ErrorAction Continue`) precisely so they keep going through partial failures - but those errors then vanish.

### Verified live on Windows 11
```powershell
$ps=[powershell]::Create()
$null=$ps.AddScript({ Write-Error 'boom'; return 7 }.ToString())
$h=$ps.BeginInvoke(); $r=$ps.EndInvoke($h)
$r              # 7   (caller gets the result...)
$ps.HadErrors   # True   (...but never checks this)
$ps.Streams.Error.Count  # 1   (errors discarded on Dispose)
$ps.Dispose()
```

### Fix
After `EndInvoke`, forward the runspace error stream to the caller's error stream (transcript-visible). PR linked. (Follow-up: the maintainer may want to filter "expected" errors such as removing an app that isn't installed - but swallowing all of them is the current bug.)

> Root cause verified live on Windows 11.

---
**Test environment:** Windows 11 25H2 (build 26200.8655), Windows PowerShell 5.1.26100.8655, .NET 4.0.30319.42000

Runspace error handling is a Windows PowerShell 5.1 + .NET 4.x behaviour (not WPF/UI), so it is build-independent - every Windows 11 build (22H2/23H2/24H2/25H2) ships PowerShell 5.1; only this build was exercised.

HetCreep:
Fixes #654

In GUI mode `Invoke-NonBlocking` runs the scriptblock in a separate runspace and called `EndInvoke` without inspecting `$ps.HadErrors` / `$ps.Streams.Error`. Every GUI-mode op writing non-terminating errors (app removal with `-ErrorAction Continue`, winget uninstall, `Disable-ScheduledTask`, `Disable-WindowsOptionalFeature`) had its errors silently discarded on `Dispose` - no feedback on failure.

This forwards the runspace error stream to the caller's error stream after `EndInvoke`:
```powershell
if ($ps.HadErrors) {
    foreach ($runspaceError in $ps.Streams.Error) {
        Write-Error -ErrorRecord $runspaceError
    }
}
```
Root cause **verified live on Windows 11** (`HadErrors=True`, `Streams.Error.Count` > 0, caller saw nothing). File parses clean.

> Note: this surfaces previously-hidden errors; the maintainer may want to filter expected ones (e.g. removing a not-installed app) as a follow-up. Swallowing all of them is the bug being fixed.

## Summary

Fixes a bug in `Invoke-NonBlocking` where errors generated during GUI mode operations are silently discarded. The function now inspects the background PowerShell runspace's error stream after `EndInvoke()` and forwards any non-terminating errors to the caller using `Write-Error`. This prevents operations like app removal, winget uninstall, scheduled task disabling, and optional feature toggling from failing silently with no user feedback.

## Changes

- **Scripts/Threading/Invoke-NonBlocking.ps1**: Added error stream inspection that checks `$ps.HadErrors` and re-emits errors from `$ps.Streams.Error` to the caller's error stream before runspace disposal (+9 lines)

---
**Test environment:** Windows 11 25H2 (build 26200.8655), Windows PowerShell 5.1.26100.8655, .NET 4.0.30319.42000

Runspace error handling is a Windows PowerShell 5.1 + .NET 4.x behaviour, build-independent across Windows 11 builds (all ship PowerShell 5.1); only this build was exercised.

### DIFF
--- a/Scripts/Threading/Invoke-NonBlocking.ps1
+++ b/Scripts/Threading/Invoke-NonBlocking.ps1
@@ -45,6 +45,15 @@ function Invoke-NonBlocking {
 
         $result = $ps.EndInvoke($handle)
 
+        # Surface non-terminating errors raised inside the runspace so GUI-mode operations
+        # (e.g. failed app removals) don't fail silently - the runspace keeps its own error
+        # stream that is otherwise discarded on Dispose.
+        if ($ps.HadErrors) {
+            foreach ($runspaceError in $ps.Streams.Error) {
+                Write-Error -ErrorRecord $runspaceError
+            }
+        }
+
         if ($result.Count -eq 0) { return $null }
         if ($result.Count -eq 1) { return $result[0] }
         return @($result)


HetCreep:
Note vs the already-merged #617 (so this isn't mistaken for a duplicate): #617 surfaces app-removal failures via `Write-Verbose` in `RemoveApps`'s per-app `catch`. This change is a **different layer** and is complementary:

- In GUI mode the actual `Remove-AppxPackage -ErrorAction Continue` / winget / `Disable-ScheduledTask` calls run inside the background runspace created by `Invoke-NonBlocking`. Their **non-terminating** errors go to that runspace's own `$ps.Streams.Error`.
- `EndInvoke` does not throw for those, so the outer `try/catch` in `RemoveApps` (the one #617 added the `Write-Verbose` to) never sees them, and they are discarded on `$ps.Dispose()`.
- This PR forwards `$ps.Streams.Error` after `EndInvoke`, so those runspace errors actually surface instead of vanishing.

So #617 covers per-app terminating exceptions at the RemoveApps layer; this covers the swallowed non-terminating errors at the threading layer. Verified live on Windows 11 25H2 (build 26200.8655, PS 5.1.26100.8655): a scriptblock that writes an error returns its result with `$ps.HadErrors = True` while the caller sees nothing.

Raphire:
Heya,

This seems like a fine addition, thanks

Issue #202, opened 2025-02-20
Uninstall failed with exit code: 2147747483


ghyujkilop:
**Describe the bug**
OneDrive Removal Attempt that didn't proceed based on geographic region.

**To Reproduce**
1. Open App Removal
2. Select Microsoft.Onedrive
3. This error log comes up:
Attempting to remove Microsoft.OneDrive...
The `msstore` source requires that you view the following agreements before using.
Terms of Transaction: https://aka.ms/microsoft-store-terms-of-transaction
The source requires the current machine's 2-letter geographic region to be sent to the backend service to function properly (ex. "US").
Found Microsoft OneDrive [Microsoft.OneDrive]
Starting package uninstall...
Uninstall failed with exit code: 2147747483

**Expected behavior**
Uninstalls like all the rest.

**Screenshots**

**Additional context**
My assumption is that it was already removed (as I tried to uninstall it via apps). But also, my PC uses a non-US geographic code. Also, installed 99 other apps alongisde this, and there was no errors. I opened Microsoft Store, and there was no terms I could see that required me to agree to them.

Raphire:
Heya,

The agreements should be automatically accepted when the script executes. The error code seems to suggest OneDrive is running and blocking removal, although I have personally also run into a situation where the uninstall seemed to error out, but it actually did succeed in the end.

Was OneDrive still installed after running the script?

ghyujkilop:
> Heya,
> 
> The agreements should be automatically accepted when the script executes. The error code seems to suggest OneDrive is running and blocking removal, although I have personally also run into a situation where the uninstall seemed to error out, but it actually did succeed in the end.
> 
> Was OneDrive still installed after running the script?

So I think it does have to do with the country code, because my country code uses UK/NZ. I believe I uninstalled OneDrive prior to running the script (this would have been manually from the installed apps settings). But I figured that, reporting this bug might be helpful for others who experience this code.

Raphire:
Interesting, it's possible winget thought it was still installed, which would explain why the uninstall failed. 

As for the country code. I don't think the country code matters in this error message. The 'US' in the output is only an example, I am from europe and have not had issues with it in the past. 

Winget (which the script uses for OneDrive removal) will by default require user agreement to sending the 2 character country code. The output you see is part of that. Normally you'd get a prompt to accept this, but the script does this automatically. Additionally, if the execution failed on sending the country code it wouldn't proceed to the uninstall phase.

Raphire:
[...]

PR #658, opened 2026-06-21
fix(app-removal): detect WinGet uninstall failures by exit code, not English text

### LINKED ISSUE #657: WinGet uninstall success is detected via English-only text; OneDrive removal failures are never checked
### Component
`Scripts/AppRemoval/RemoveApps.ps1` (the WinGet uninstall path, used for OneDrive and Microsoft Edge)

### Summary
The success/failure of `winget uninstall` is inferred by string-matching **English** console output, and for **OneDrive** the result is never checked at all. Two failure modes result:

**1. Locale-dependent detection (every WinGet removal).**
Failure is detected with:
```powershell
$wingetFailed = Select-String -InputObject $wingetOutput -Pattern "Uninstall failed with exit code|No installed package found matching input criteria|No package found matching input criteria"
```
Those are English strings. On a non-English Windows, `winget` prints localized text, so the patterns never match and a failed uninstall is treated as success. For Edge this sets `$edgeUninstallSucceeded = $true` and the "force uninstall" fallback is never offered, even though Edge was not removed.

**2. OneDrive result is discarded (silent failure).**
`$wingetFailed` is only consumed inside the `if ($isEdgeId)` block. OneDrive (`Microsoft.OneDrive`) is removed via WinGet too, but it is not an Edge id, so the block is skipped and the loop hits `continue`. If the OneDrive uninstall fails, nothing is printed and the user gets no indication.

### Why it matters
- Non-English Windows is a large share of installs; on those systems Edge/OneDrive removal can report success while leaving the app in place.
- OneDrive removal failures are silent on every locale.

### Root cause
`winget` reports success/failure through its **exit code**, which is locale-independent (see [WinGet exit codes](https://learn.microsoft.com/windows/package-manager/winget/troubleshooting#exit-codes)). The current code ignores the exit code and parses localized text. The exit code is lost because the `winget` call runs inside an `Invoke-NonBlocking` background runspace and only the text output is returned.

### Suggested fix
Capture `$LASTEXITCODE` inside the runspace scriptblock and treat any non-zero code (or a null result) as failure, then apply that check to OneDrive as well as Edge. PR to follow.

### Environment
Reviewed against `master` @ `8289417`. Code analysis only -- the WinGet uninstall path is destructive, so it was not executed; winget exit-code behavior confirmed from Microsoft Learn. Prior issue #202 ("Uninstall failed with exit code: 2147747483") shows winget surfacing a real non-zero exit code.

HetCreep:
## Problem
Fixes #657. The WinGet uninstall path in `Scripts/AppRemoval/RemoveApps.ps1` judges success by string-matching English console text, and never checks the result for OneDrive.

## Two bugs, one root cause
**Locale-dependent detection.** `$wingetFailed` was derived from a `Select-String` match on `"Uninstall failed with exit code| No installed package found... | No package found..."`. On non-English Windows the winget output is localized, the patterns miss, and a failed uninstall reads as success. For Edge that set `$edgeUninstallSucceeded = $true` and the force-uninstall fallback was never offered.

**OneDrive result discarded.** `$wingetFailed` was only used inside `if ($isEdgeId)`. OneDrive is removed via WinGet too but is not an Edge id, so its failure was silently ignored on every locale.

## Fix
winget reports success/failure through its exit code, which is locale-independent ([docs](https://learn.microsoft.com/windows/package-manager/winget/troubleshooting#exit-codes)). The exit code was being thrown away because the `winget` call runs inside an `Invoke-NonBlocking` runspace and only the text output was returned.

- Capture `$LASTEXITCODE` inside the scriptblock and return it alongside the output.
- `$wingetFailed = ($null -eq $wingetResult) -or ($wingetResult.ExitCode -ne 0)` -- a null result (timeout / not run) or any non-zero code is a failure.
- Add an `elseif ($wingetFailed)` for the OneDrive case so its failures are surfaced (`Unable to uninstall <app> via WinGet`).

## Behavior change
- Edge: force-uninstall fallback now triggers on real failures regardless of UI language (previously English-only).
- OneDrive: a failed uninstall now prints an error instead of nothing.
- Success path and the actual removal mechanism are unchanged.

## Verification
- Reviewed against `master` @ `8289417`; file is parse-clean via `[Parser]::ParseFile`.
- winget exit-code behavior confirmed from Microsoft Learn; prior issue #202 shows winget surfacing a real non-zero exit code.
- The WinGet uninstall path is destructive, so it was not executed live -- this is a code-level fix verified by inspection and parse.

## Summary

Fixed two related WinGet uninstall issues in `Scripts/AppRemoval/RemoveApps.ps1` that caused uninstall failures to be misreported as success on non-English Windows and prevented OneDrive uninstall failures from being surfaced.

## Changes

Updated `Scripts/AppRemoval/RemoveApps.ps1` to determine WinGet uninstall success/failure using WinGet’s locale-independent process exit code instead of matching English console output:

- Replaced locale-dependent `Select-String` matching of English uninstall failure text with exit-code checking
- Updated the `Invoke-NonBlocking` scriptblock to capture `$LASTEXITCODE` and return it (alongside output) so it isn’t lost
- Changed failure detection to: `$wingetFailed = ($null -eq $wingetResult) -or ($wingetResult.ExitCode -ne 0)`
- Added an additional failure-handling path (`elseif ($wingetFailed)`) for non-Edge WinGet removals (including OneDrive) so uninstall failures emit an explicit red error message instead of being silently ignored
- Ensured the WinGet uninstall path includes OneDrive/Edge and the additional Edge app-id `XP9CXNGPPJ97XX`

Edge force-uninstall fallback is now correctly triggered on failed uninstalls across Windows locales since it no longer depends on localized console text.

## Impact

- **Locale-independent detection**: Uninstall failure is now determined by WinGet’s exit code (0 success, non-zero failure), avoiding false “success” reports on non-English systems
- **Edge force-uninstall fallback restored**: Failed Edge uninstalls correctly trigger the force-uninstall fallback regardless of Windows language
- **OneDrive failure visibility**: OneDrive removal failures are no longer silently ignored; users receive error messaging when WinGet uninstall fails

## Lines changed

+11/-3

### DIFF
--- a/Scripts/AppRemoval/RemoveApps.ps1
+++ b/Scripts/AppRemoval/RemoveApps.ps1
@@ -82,12 +82,16 @@ function RemoveApps {
             }
             else {
                 # Uninstall app via WinGet
-                $wingetOutput = Invoke-NonBlocking -ScriptBlock {
+                $wingetResult = Invoke-NonBlocking -ScriptBlock {
                     param($appId)
-                    winget uninstall --accept-source-agreements --disable-interactivity --id $appId
+                    $output = winget uninstall --accept-source-agreements --disable-interactivity --id $appId
+                    [PSCustomObject]@{ ExitCode = $LASTEXITCODE; Output = $output }
                 } -ArgumentList $app
 
-                $wingetFailed = Select-String -InputObject $wingetOutput -Pattern "Uninstall failed with exit code|No installed package found matching input criteria|No package found matching input criteria" -SimpleMatch:$false
+                # winget reports success/failure via its exit code, which is locale-independent.
+                # The previous match on English console text silently passed on non-English Windows.
+                # Treat a null result (timed out / not run) or any non-zero exit code as a failure.
+                $wingetFailed = ($null -eq $wingetResult) -or ($wingetResult.ExitCode -ne 0)
                 if ($isEdgeId) {
                     if (-not $wingetFailed) {
                         $edgeUninstallSucceeded = $true
@@ -116,6 +120,11 @@ function RemoveApps {
                             ForceRemoveEdge
                         }
                     }
+                }
+                elseif ($wingetFailed) {
+                    # OneDrive is the only other app removed via WinGet; surface the failure
+                    # instead of silently continuing (previously its result was never checked).
+                    Write-Host "Unable to uninstall $app via WinGet" -ForegroundColor Red
                 }
             }
 
--- a/Scripts/AppRemoval/RemoveApps.ps1
+++ b/Scripts/AppRemoval/RemoveApps.ps1
@@ -84,6 +84,7 @@ function RemoveApps {
                 # Uninstall app via WinGet
                 $wingetResult = Invoke-NonBlocking -ScriptBlock {
                     param($appId)
+                    $global:LASTEXITCODE = 0
                     $output = winget uninstall --accept-source-agreements --disable-interactivity --id $appId
                     [PSCustomObject]@{ ExitCode = $LASTEXITCODE; Output = $output }
                 } -ArgumentList $app
@@ -122,8 +123,6 @@ function RemoveApps {
                     }
                 }
                 elseif ($wingetFailed) {
-                    # OneDrive is the only other app removed via WinGet; surface the failure
-                    # instead of silently continuing (previously its result was never checked).
                     Write-Host "Unable to uninstall $app via WinGet" -ForegroundColor Red
                 }
             }


Raphire:
While this is a better approach, unfortunately WinGet often returns an error even after a successful uninstall. This is at least the case for OneDrive as well as the new Copilot app (Which I added support for in #662). This was already the case before your changes though.

I will merge this for now, but I may need to look into changing the messaging of the error to indicate that Uninstall may have succeeded regardless of the WinGet output.

HetCreep:
Thanks for merging! And good catch -- you're right that winget's output is unreliable in both directions (false failures on successful uninstalls like OneDrive/Copilot, plus the localized-text issue this PR addressed). The exit code mainly fixes the non-English case; it doesn't make winget's result trustworthy on its own.

For the messaging rework you mentioned: one option that avoids trusting winget at all is to confirm success by checking whether the package is actually still present afterwards (e.g. `Get-AppxPackage` / the OneDrive install path) and treat winget's own output as advisory. Happy to help if that's useful -- otherwise I'll leave it to your work in #662.

Issue #675, opened 2026-06-25
Command Excuting error? (Windows 10)


Kate-alt-69:
### Checklist

- [x] I have searched for existing issues/discussions and didn't find any similar ones.
- [x] I haven't used any other scripts, tools or programs that might have caused this issue.

### Windows version

Windows 10 HSL 22H2 Installed on 26-‎09-‎2025 OS Build 19045.7417

### Script mode/options

custom mode

### Describe the issue

```powershell



                         ^
                        / \
                       /   \
                      /     \
                     / ===== \
                     |  ---  |
                     | ( O ) |
                     |  ---  |
                     |       |
                    /|       |\
                   / |       | \
                     |  '''  |    *
                       (''')    *  *
                       ( ' )   *

             Win11Debloat is launching...
                Keep this window open


Retrieving installed apps via winget...
<ERROR#2>
<ERROR#2>
Selected user mode: current user (Merge)
> Creating registry backup...
Backup successfully created: C:\Users\Merge\Downloads\Win11Debloat-master\Backups\Win11Debloat-RegistryBackup-20260625_223942.json

> Creating a system restore point...
Error: Unable to retrieve existing restore points: The term 'Get-ComputerRestorePoint' is not recognized as a name of a cmdlet, function, script file, or executable program.
Check the spelling of the name, or if a path was included, verify that the path is correct and try again.
Warning: Continuing without restore point

> Removing selected apps for all users...
9 apps selected for removal
Removing Microsoft.OutlookForWindows
<ERROR#1>
Removing Microsoft.MSPaint
<ERROR#1>
Removing Microsoft.YourPhone
<ERROR#1>
Removing Microsoft.Windows.Photos
<ERROR#1>
Removing Microsoft.SkypeApp
<ERROR#1>
Removing Microsoft.WindowsSoundRecorder
<ERROR#1>
Removing Microsoft.WindowsMaps
<ERROR#1>
Removing Microsoft.XboxApp
<ERROR#1>
Removing Microsoft.XboxSpeechToTextOverlay
<ERROR#1>
> Disabling telemetry and diagnostic data collection...
The operation completed successfully.

> Disabling telemetry scheduled tasks...

> Disabling tips & tricks on the lock screen...
The operation completed successfully.

> Preventing automatic restarts after updates while signed in...
The operation completed successfully.

> Hiding the chat icon from the taskbar...
The operation completed successfully.

> Enabling file extensions for known file types...
The operation completed successfully.

[...]

PR #676, opened 2026-06-26
fix(startup): refuse to run under PowerShell 7 with a clear message

HetCreep:
## Problem
Refs #675. Running Win11Debloat under **PowerShell 7** (pwsh) breaks: the Appx module (`Get-AppxPackage` / `Remove-AppxPackage`) fails to load with *"Operation is not supported on this platform"* (`0x80131539`), and `Get-ComputerRestorePoint` doesn't exist there. So app removal and restore points silently fail -- and the run still prints *"All changes have been applied successfully!"*, leaving the user thinking it worked when nothing was removed.

Both are Windows-PowerShell-5.1-only APIs; PS7 needs the WinPS compatibility shim (a known PowerShell limitation -- PowerShell/PowerShell#13138, #14057).

## Fix
A guard at startup, placed **before the admin check** so it also catches the already-elevated case from #675: if running under the Core edition (pwsh), print a clear message to re-run with `powershell.exe` and exit, instead of failing mid-run.

## Verification
- Parse-clean (`[Parser]::ParseFile`).
- On Windows PowerShell 5.1 (Desktop edition) the guard condition is `$false`, so it doesn't affect normal runs -- verified on `5.1.26100.8737`.
- I couldn't exercise the PS7 trigger path on my end (no pwsh session here), so worth a quick check under PowerShell 7 before merge.

## Notes
- This redirects to Windows PowerShell 5.1 rather than trying to make WD run under PS7 -- 5.1 ships in-box permanently, so redirecting is the right level of support.
- Separately (not in this PR): the *"applied successfully"* message in `Show-ApplyModal.ps1` is printed unconditionally even when removals fail -- flagged on #675 as a broader observability issue.

Added an early startup guard that detects when Win11Debloat is running under PowerShell 7/Core (before the admin check), prints a clear message instructing the user to rerun it with `powershell.exe`, and exits with code `1`. This prevents the script from continuing in an unsupported environment where Windows PowerShell 5.1-only APIs (e.g., `Get-AppxPackage`/`Remove-AppxPackage` and `Get-ComputerRestorePoint`) can fail or behave silently while still reporting success.

### DIFF
--- a/Win11Debloat.ps1
+++ b/Win11Debloat.ps1
@@ -99,6 +99,17 @@ param (
     [switch]$HideDriveLetters
 )
 
+# Win11Debloat depends on Windows PowerShell 5.1 cmdlets (the Appx module's Get-AppxPackage /
+# Remove-AppxPackage, and Get-ComputerRestorePoint) that do not load in PowerShell 7 (pwsh), where the
+# Appx module fails with "Operation is not supported on this platform" (0x80131539). Without this guard
+# the run continues and silently fails to remove any apps while still reporting success. See issue #675.
+if ($PSVersionTable.PSEdition -eq 'Core') {
+    Write-Host "Win11Debloat requires Windows PowerShell 5.1, but it is running under PowerShell $($PSVersionTable.PSVersion) (pwsh / Core edition)." -ForegroundColor Red
+    Write-Host "App removal and system restore points rely on modules that are not available in PowerShell 7, so the run cannot complete correctly here." -ForegroundColor Red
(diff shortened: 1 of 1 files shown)


zoicware:
i came to the same conclusion with my script while iirc you can import the 5.1 appx module and dism if the user is running 7 but after so many issues and fixes i just gave up and prevented the script from running with pwsh

interestingly after lots of confusion what version the user was running it seems ms has updated the default 5.1 header to clearly state what version and no longer links to download pwsh 7 😄 
D992D599-388E-4E90-93A0-016323B5A744

PR #685, opened 2026-06-30
fix(restore-gui): surface registry-backup load failures to the user

HetCreep:
Picking a malformed or invalid registry backup .json in the Restore Backup dialog throws out of the WPF button-click handler (Load-RegistryBackupFromFile throws on bad JSON or schema-validation failure; the overview render throws on a backup with no restorable changes). Because the throw originates inside a WPF dispatcher callback, it bypasses every surrounding try/catch -- the dialog's own ShowDialog try/catch and Show-RestoreBackupWindow's try/catch both sit on a different call stack than the dispatcher-invoked click handler -- and is caught only by Show-MainWindow's global Dispatcher.UnhandledException handler, which only writes a console warning (invisible in a windowed GUI) and sets Handled=true. Net effect: the user picks a bad file and the dialog silently does nothing -- no error, no feedback, it just sits there. This is the most-reachable restore failure: one wrong pick in the file-open dialog (a non-backup json, a truncated download, a backup for another machine). FIX: wrap the load + overview render in try/catch and surface the failure via the existing Show-MessageBox helper -- the same pattern already used elsewhere in this file. The dialog stays open so the user can pick a different file. VERIFICATION: parse-clean; the Show-MessageBox call matches the existing usage signature in this file (-Owner -Title -Message -Button OK -Icon). I did not launch the full WPF GUI end-to-end this session; the change is a localized try/catch around the two throwing calls using the file's own error-dialog helper. Cross-checked against Microsoft's documented Dispatcher.UnhandledException interception-before-ShowDialog-try/catch behavior, so the wrapping is inside the click handler where it can actually catch the throw.

This change makes restore-backup failures visible to users instead of failing silently in the WPF restore dialog.

- Wraps backup loading and registry overview rendering/initialization in a local `try/catch` in `Show-RestoreBackupDialog.ps1` (so exceptions don’t escape the click handler)
- Shows an “Invalid Backup File” message via the existing `Show-MessageBox` helper, including `($_.Exception.Message)`, and keeps the dialog open so the user can choose another backup file
- Updates `Normalize-RegistryBackup` to wrap allow-list validation in a `try/catch`, appending `Failed to validate backup: <exception message>` to `$errors` so runtime validation exceptions are captured and then handled by the existing aggregation/throw logic

### DIFF
--- a/Scripts/Features/RestoreRegistryBackup.ps1
+++ b/Scripts/Features/RestoreRegistryBackup.ps1
@@ -1,3 +1,19 @@
+<#
+    .SYNOPSIS
+        Loads a registry backup from a JSON file and normalizes its contents.
+
+    .DESCRIPTION
+        Loads a registry backup from disk and returns a normalized representation
+        of its contents suitable for use by the restore workflow. Throws if the
+        file is missing, unreadable, or not valid JSON.
+
+    .PARAMETER FilePath
+        The absolute path to the registry backup JSON file to load.
+
+    .OUTPUTS
+        PSCustomObject
+        A normalized registry backup object produced by Normalize-RegistryBackup.
+#>
 function Load-RegistryBackupFromFile {
     param(
         [Parameter(Mandatory)]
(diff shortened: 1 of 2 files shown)


Raphire:
Heya,

Thanks for flagging this, I have made some additional changes to ensure the errors in Normalize-RegistryBackup are properly  aggregated and logged. I also added docstrings which is something that I try to add to all functions touched in PR's going forward.

PR #700, opened 2026-07-13
fix: honor cancellation during changes

ShiroKSH:
## Summary

- Stop CLI execution cleanly after a restore-point failure is cancelled, without restarting Explorer or reporting success.
- Honor cancellation between telemetry scheduled-task operations and deferred Edge uninstall steps.
- Only create target-user RunOnce uninstall tasks after the active WinGet operation completes without cancellation.

## Why

The GUI pumps events while long-running operations execute. A cancellation request could previously allow later telemetry and Edge operations to continue. In the CLI restore-point flow, declining to continue still restarted Explorer and printed a successful completion message.

## Validation

- Parsed every PowerShell script with the Windows PowerShell parser.
- Parsed all configuration JSON and checked the staged diff for whitespace errors.
- Ran mocked cancellation checks for telemetry, Edge removal, post-removal verification, and target-user RunOnce scheduling.

## PR summary

- Tightens cancellation handling across the main flow, system restore-point creation, app removal, telemetry scheduled tasks, and deferred uninstall/Edge force-remove steps.
- Stops CLI/normal execution cleanly when restore-point creation fails and the user cancels continuation (via `$script:CancelRequested`), preventing any further change phases from running.
- Makes completion behavior conditional on cancellation: after `Invoke-AllChanges`, the script warns about remaining changes and exits via `AwaitKeyToExit` when `$script:CancelRequested` is set, avoiding Explorer restart and success reporting.
- Updates `Disable-TelemetryScheduledTasks` / `Enable-TelemetryScheduledTasks` to bail out immediately during cancellation by exiting early inside per-task `foreach` loops before enable/disable logic runs.
- Reworks `RemoveApps` to process Microsoft Edge WinGet app IDs directly in the main per-app loop (removing the prior `edgeAppsInList`/deferred Edge flow), and to exit immediately on `$script:CancelRequested` before any post-removal verification or Edge force-remove prompting.
- Refines `RemoveApps` post-removal verification to run only for `wingetRemovedApps`, with per-app installed checks; for Edge it calls `Request-EdgeForceRemove` only once, and otherwise reports generic WinGet uninstall failures.
- Updates `Remove-WinGetApp` so `winget uninstall` is invoked first, and RunOnce uninstall registry tasks for the configured `User`/`Sysprep` scope are scheduled only after WinGet completes successfully without cancellation.
- Includes PowerShell cleanup (syntax/whitespace), JSON/whitespace adjustments, and mocked cancellation validation to cover the updated cancellation behavior.

### DIFF
--- a/Scripts/AppRemoval/RemoveApps.ps1
+++ b/Scripts/AppRemoval/RemoveApps.ps1
@@ -73,6 +73,11 @@ function RemoveApps {
         Remove-EdgeApp -edgeAppsInList $edgeAppsInList
     }
 
+    if ($script:CancelRequested) {
+        Write-Host ""
+        return
+    }
+
     # Check whether any winget-removed apps are still present, and report errors for each one.
     if ($wingetRemovedApps.Count -gt 0 -or $edgeAppsInList.Count -gt 0) {
         $postRemovalList = if ($script:WingetInstalled) { GetInstalledAppsViaWinget -TimeOut 10 -NonBlocking } else { $null }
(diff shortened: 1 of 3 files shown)


Raphire:
Heya,

Thanks for taking the time to contribute.  While reviewing your code I found some redundant logic around the edge removal so I refactored that & included it here.

PR #702, opened 2026-07-14
fix: escape Windows Terminal path

ShiroKSH:
## Summary

- Escape the Windows Terminal executable path before embedding it in the launcher PowerShell command.

## Root cause

An apostrophe in a Windows profile path produced invalid PowerShell before elevation.

## Validation

- Parsed the launcher command with apostrophe and ordinary profile paths.
- Parsed all PowerShell source files, configuration JSON, and primary XAML.

Escapes the Windows Terminal executable path before embedding it in the launcher PowerShell command, preventing failures when Windows profile paths contain apostrophes. Validation covers apostrophe-containing and standard paths, PowerShell sources, configuration JSON, and primary XAML.


muhammadameer01-boop (approved):

STiFLeR7:
Verified the fix against the actual source of \`wtPath\`: it's set from \`wtDefaultPath\` (\`%LOCALAPPDATA%\Microsoft\WindowsApps\wt.exe\`) or \`wtScoopPath\` (\`%USERPROFILE%\scoop\apps\...\`), both of which expand through \`%LOCALAPPDATA%\`/\`%USERPROFILE%\` — so this isn't a hypothetical edge case, it reproduces for any real Windows account whose username contains an apostrophe (e.g. \`C:\Users\O'Brien\...\`), since that apostrophe flows straight into \`wtPath\`.

The fix correctly mirrors the escaping already used for \`$p='%SCRIPT_PATH:'=''%'\` two tokens earlier in the same line — \`%wtPath:'=''%\` doubles any embedded single quote via batch string substitution, which is exactly PowerShell's own single-quoted-string escape convention. Consistent with the existing pattern rather than introducing a new one. Looks correct to me.

Raphire:
Heya,

Thanks for taking the time to contribute. This change seems to do the trick in properly parsing paths with uncommon characters. Good stuff.

PR #643, opened 2026-06-21
feat: Add dynamic multi-language support for the GUI

jmolinap2:
## Summary

This PR adds a dynamic localization system to the Win11Debloat GUI, allowing the interface to display in the user's system language automatically.

### Changes

- **`Scripts/FileIO/LoadLanguage.ps1`** — new language loader: detects system UI culture, loads the matching JSON file from `Config/Languages/`, falls back to `en-US`. Provides helper functions: `L`, `Get-LangCategory`, `Get-LangFeature`, `Get-LangGroupLabel`, `Get-LangGroupValue`, `Get-LangFeatureTooltip`, `Get-LangGroupTooltip`
- **`Config/Languages/en-US.json`** — full English strings for all UI labels, tooltips, categories, messages and feature descriptions
- **`Config/Languages/es-ES.json`** — complete Spanish (Castilian) translation
- **`Schemas/MainWindow.xaml`** — all hardcoded strings replaced with `%LANG:key%` markers
- **`Scripts/GUI/`** — all GUI scripts updated to use `L 'key'` calls and lang helper functions for dynamically-built controls (tweaks, app list, tooltips)
- **`Win11Debloat.ps1`** — loads language on startup via `$script:Lang = LoadLanguage`

### How it works

On startup, the script reads `[System.Globalization.CultureInfo]::CurrentUICulture`. It tries an exact match (`es-ES.json`), then a prefix match (`es-*.json`), then falls back to `en-US.json`. New languages can be added by dropping a JSON file in `Config/Languages/`.

### Adding a new language

Create `Config/Languages/<culture-code>.json` following the structure of `en-US.json`. The loader will pick it up automatically.

## Multi-Language Localization System Implementation

This PR introduces a comprehensive dynamic localization framework to the Win11Debloat GUI, enabling the interface to automatically display in the user's system language with a fallback to English.

### Core Infrastructure

**`Scripts/FileIO/LoadLanguage.ps1`** — New module providing language detection and localization helpers:
- `LoadLanguage` function: Detects the system UI culture (e.g., `es-ES`, `en-US`) and loads the corresponding JSON language file from `Config/Languages/`. Uses an "exact match → prefix glob → `en-US` fallback" strategy
- `ConvertTo-LocalizedXaml` function: Substitutes `%LANG:Key%` markers in XAML with translated values, XML-escaping results for safe XAML attribute usage
[...]

Issue #499, opened 2026-03-05
Multilingual localization


Lintern:
### Checklist

- [x] I have searched for existing issues/discussions and didn't find any similar ones.

### Describe the feature or improvement

Hopefully the UI interface can be multilingual

Mikachu2333:
need it too

Raphire:
Heya,

I don't have any ETA, but this is something high on my priority list.

Zarijaden:
We need Chinese, Really! Thanks really much.

Sqidy-P:
Hey there,

My first language is Spanish--This is a feature I'd like to work on adding.

I'll see what I can do!

IntelAMDNVIDIA:
**① 开场：呼应 #499 现状**

中文：@Raphire 看到你在 #499 里说多语言是"高优先级"但还没排期，社区也很想要——@Zarijaden 明确说急需中文，@Sqidy-P 也表示想来做西班牙语版。我想在动手前先和你对齐，避免重复造轮子。

English: @Raphire I see in #499 you marked multilingual as "high priority" but without an ETA, and the community clearly wants it — @Zarijaden explicitly asked for Chinese, and @Sqidy-P offered to work on the Spanish version. I'd like to align with you before starting, to avoid duplicate effort.

---

**② 协调：已有两路人马，需统一**

中文：目前看起来已有两条相关线索：@jmolinap2 的 PR #643（GUI 动态多语言，en-US + es-ES）和 @Sqidy-P 的表态。我倾向于接手/协调这些努力，而不是再开一个竞争的 PR。你希望社区以什么方式参与——是各自提 PR，还是有人牵头做统一实现？

English: There seem to be two related threads already: PR #643 by @jmolinap2 (dynamic GUI localization, en-US + es-ES) and @Sqidy-P's offer. I'd prefer to coordinate / take over these efforts rather than open yet another competing PR. How would you like the community to engage — separate PRs, or one person leading a unified implementation?

---

**③ 我们的目标：模块化 i18n 框架**

中文：我的具体目标是做一个模块化 i18n 框架：
- `en-US` 作为唯一事实来源（默认 + 回退），其他语言都只是数据；
- 覆盖范围不止 GUI：控制台/CLI 文本、`Config/*.json` 的展示标签（Features/Apps）、以及 XAML，全部走同一个翻译层；
- 用户能主动选择语言（CLI `-Language` 参数 + GUI 下拉），而不只是按系统区域自动检测；
[...]

PR #715, opened 2026-07-24
fix: support QWord, ExpandString and MultiString in the .reg fallback writer

SashankBhamidi:
So, I was walking a friend through the PowerShell fallback in Import-RegistryFile.ps1, using this repo as an example of when `reg import` fails and you need a backup writer. I noticed that Convert-RegOperationToValueKind only handles DWord, String and Binary, everything else throws "Unsupported value type".

It's not a rare fallback either. Import-RegistryFile.ps1 skips `reg import` entirely whenever `-User` targets someone already logged in (their hive's mounted under its own SID, not `HKEY_USERS\Default`), and it's the retry path whenever `reg import` fails. There's no shipped .reg file that uses QWord/hex(2)/hex(7) today, but the next feature that needs one would hit this silently.

I added the three missing cases, swapped a duplicated value-name normalization for the existing helper too.

@Raphire didn't spot anything pointing to this being on purpose.

## Summary by CodeRabbit

- Expanded `.reg` fallback registry conversion support for `QWord`, `ExpandString`, and `MultiString` values.
- Reused `Get-NormalizedRegistryValueName` for value-name normalization.
- Added test coverage for the newly supported registry value types.

### DIFF
--- a/Scripts/Helpers/Apply-RegistryRegFile.ps1
+++ b/Scripts/Helpers/Apply-RegistryRegFile.ps1
@@ -11,6 +11,10 @@ function Get-NormalizedRegistryValueName {
     return [string]$ValueName
 }
 
+<#
+    .SYNOPSIS
+        Converts a parsed .reg operation into a Name/Kind/Value set for RegistryKey.SetValue.
+#>
 function Convert-RegOperationToValueKind {
     param(
         [Parameter(Mandatory)]
--- a/Scripts/Helpers/Apply-RegistryRegFile.ps1
+++ b/Scripts/Helpers/Apply-RegistryRegFile.ps1
@@ -17,21 +17,34 @@ function Convert-RegOperationToValueKind {
         $Operation
     )
 
-    $valueName = if ([string]::IsNullOrEmpty([string]$Operation.ValueName)) { '' } else { [string]$Operation.ValueName }
+    $valueName = Get-NormalizedRegistryValueName -ValueName $Operation.ValueName
     $valueType = [string]$Operation.ValueType
     $operationKeyPath = [string]$Operation.KeyPath
 
+    # ValueType here is whatever Get-RegFileOperations parsed it as.
+    # Hex2/Hex7 are its names for REG_EXPAND_SZ/REG_MULTI_SZ, already decoded to string/string[].
     switch ($valueType) {
         'DWord' {
             $unsigned = [uint32]$Operation.ValueData
             $value = [BitConverter]::ToInt32([BitConverter]::GetBytes($unsigned), 0)
             return @{ Name = $valueName; Kind = [Microsoft.Win32.RegistryValueKind]::DWord; Value = $value }
         }
+        'QWord' {
+            $unsigned = [uint64]$Operation.ValueData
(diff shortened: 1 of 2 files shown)


Raphire:
Heya,

Thanks for taking the time to contribute again. I did previously neglect to (fully) add support for these value kinds as there was no explicit need for it, but this could indeed cause issues when a setting does need to use any of these value kinds.

SashankBhamidi:
@Raphire,

No worries, makes sense it just never came up. Thanks for merging!
Also I have some time off coming up if you could use an extra hand on anything, happy to help with issues/PRs or take on something specific if you've got a priority list :)

Raphire:
@SashankBhamidi That'd be most welcome, yea.

I think localization is the biggest priority right now. We do have an open PR for that (#643), but I haven't had time to really review it yet and it has got some conflicts now due to recent refactoring efforts. It would be great if you could take a look at it or experiment with it a bit on your end.

Other than that I think the most meaningful addition would be the option to apply changes to all users. The best solution would probably be to apply tweaks using machine policies, as having to loop through all user accounts to apply tweaks seems like a problem waiting to happen on machines with a large nr of users or domain joined machines. This would probably be quite a bit of research though, finding all matching policies for the current feature set.

If you're looking for some smaller features to focus on I'd probably focus on some of these new features/changes coming to Windows 11 soon that may have an impact on some existing features in Win11Debloat as well: https://www.windowscentral.com/microsoft/windows-11/7-features-im-most-excited-to-use-on-windows-11-version-26h2

SashankBhamidi:
@Raphire Hey, read through #643 since you pointed me at it. It's got merge conflicts from the refactor already, and a few things in the code itself I'd want to fix rather than build on top of as is.

`LoadLanguage.ps1` has basically the same lookup written out five separate times (`Get-LangCategory`, `Get-LangFeature`, `Get-LangGroupLabel`, and two more), just swapping which property it reads each time, that collapses into one function easily. Bigger thing is how it keys category translations, it uses the English category name itself as the key, so the moment that name changes in `Features.json` the translation quietly stops matching and you get English sitting inside an otherwise translated screen. Also a few strings use `{0}` style number formatting, which breaks for languages with real plural rules, worth getting right from the start so it's easier for people to actually contribute translations down the line, myself included, I speak a few languages and would like to help translate once this is in a good place.

No tests on the loading/substitution logic either, and it's GUI only, CLI's still fully English.

The XAML substitution idea itself holds up fine though. I'd cherry pick what's usable from jmolinap2's commits and credit the PR, but rebuild the schema properly on current master rather than patch this one. Only thing is giving categories a real ID instead of keying off the name touches `Features.json`, so wanted your take before going there.

Also want to keep everything project related here on GitHub so there's a proper log to point back to later, but if you're on Discord or Matrix or somewhere I could DM you a link to a comment when something needs a faster response, up to you.

Raphire:
@SashankBhamidi some fair points. It makes sense to give categories an ID for the reasons you mentioned.

I have recently made a discord server, to allow for better communication with/between contributers, users, etc. It's not fully fledged out yet, but you can join it here: https://discord.gg/5PhSzXEDyH

P.S. Maybe we should continue this thread in #499

PR #710, opened 2026-07-19
Improve registry backup safety and add optional backup skipping

Raphire:
This should address the concerns raised in #686

## Summary by CodeRabbit

- Added a “registry backup (Recommended)” checkbox to the main deployment UI and wired it through apply/export/import via the `SkipRegistryBackup` flag (unchecked => `SkipRegistryBackup=true`, checked => `SkipRegistryBackup=false`), including updated options text and import persistence.
- Extended the CLI/script surface to accept `-SkipRegistryBackup` and pass it through to `Win11Debloat.ps1`, plus updated `Import-ConfigToParams` to read `SkipRegistryBackup` from exported configs.
- Updated registry-backup orchestration in `Invoke-AllChanges` so the registry-backup phase (and step counting) runs only when registry backup is needed and `SkipRegistryBackup` is not set.
- Improved backup safety and restore correctness: `Convert-RegistryKeyToSnapshot` now throws when an included child subkey can’t be opened; registry restore adds a full snapshot preflight check and restores more safely by enforcing direct parent→child relationships and avoiding the prior “force full tree” deletion behavior.
- Strengthened test coverage for `SkipRegistryBackup` and backup/restore safety (including unreadable subkey snapshot failure and new snapshot preflight validation), plus added new exported config/test data for skipping registry backup.
- Minor JSON formatting touch-up in `Config/DefaultSettings.json` (re-emitted the final closing-brace line).

PR #718, opened 2026-07-25
fix: update renamed store-suggestions function calls missed in refactor

### LINKED ISSUE #717: DisableStoreSearchSuggestionsForAllUsers is not recognized
### Checklist

- [x] I have searched for existing issues/discussions and didn't find any similar ones.
- [x] I haven't used any other scripts, tools or programs that might have caused this issue.

### Windows version

Windows 11 Pro  26H1 28000.2525

### Script mode/options

Custom Mode

### Describe the issue

Edition	Windows 11 Pro
Version	26H1
Installed on	‎29.‎06.‎2026
OS build	28000.2525
Experience	Windows Feature Experience Pack 1000.26100.316.0

Image

### Steps to reproduce

I would probably try to fix issue if I could remember what I selected before hitting that last green button
and you dont seem to keep the failed selections in the custom template for a re-apply

### Error output

Image

### Additional context

_No response_

HetCreep:
The approved-verbs refactor (#708) renamed the store-suggestions all-users functions to Set-StoreSearchSuggestionsDisabledForAllUsers / Set-StoreSearchSuggestionsEnabledForAllUsers, but the two call sites in Invoke-Changes.ps1 (the Sysprep apply path and the all-users undo path) still call the OLD names -- so that path now dies with 'The term DisableStoreSearchSuggestionsForAllUsers is not recognized', exactly as reported in #717. Notably the Pester suite did not catch this because Tests/Invoke-Changes.Tests.ps1's BeforeAll stubs also defined the old names, which masked the dangling calls -- the suite passed while the production path was broken. This PR updates: the two call sites in Invoke-Changes.ps1, the test stubs/mocks/assertions to the new names (so the suite now exercises the real contract), and two stale .EXAMPLE doc lines in Set-StoreSearchSuggestions.ps1. Verification: parse-clean on all three files; grep confirms zero references to the old names remain anywhere in Scripts/ or Tests/. I could not run the Pester suite locally (in-box Pester 3.4; the suite needs v5 and I avoid installing tools on this machine) -- tests.yml should exercise it here. Fixes #717

## Summary by CodeRabbit

- Updated Store search-suggestions function references in apply and undo flows.
- Updated related Pester mocks, stubs, and assertions.
- Corrected stale documentation examples.
- Restores all-users Store search-suggestions functionality during Windows 11 Sysprep operations.

### DIFF
--- a/Scripts/Features/Invoke-Changes.ps1
+++ b/Scripts/Features/Invoke-Changes.ps1
@@ -121,7 +121,7 @@ function Invoke-FeatureApply {
         'DisableStoreSearchSuggestions' {
             if ($script:Params.ContainsKey("Sysprep")) {
                 Write-Host "> Disabling Microsoft Store search suggestions in the start menu for all users..."
-                DisableStoreSearchSuggestionsForAllUsers
+                Set-StoreSearchSuggestionsDisabledForAllUsers
                 Write-Host ""
                 return
             }
@@ -159,7 +159,7 @@ function Invoke-FeatureUndo {
         'DisableStoreSearchSuggestions' {
             if ($script:Params.ContainsKey('Sysprep')) {
                 Write-Host "> Re-enabling Microsoft Store search suggestions in the start menu for all users..."
-                EnableStoreSearchSuggestionsForAllUsers
+                Set-StoreSearchSuggestionsEnabledForAllUsers
                 Write-Host ""
                 return
             }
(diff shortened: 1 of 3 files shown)


Raphire:
Heya,

This was actually already fixed in https://github.com/Raphire/Win11Debloat/pull/710, I just wasn't able to merge it in until now. Thanks for taking the time to fix it though.

PR #724, opened 2026-07-29
fix: remove Mark-of-the-Web from script files at startup

### LINKED ISSUE #720: Add Unblock-File step to README for ZIP users (prevents GUI launch)
### Checklist

- [x] I have searched for existing issues/discussions and didn't find any similar ones.
- [x] I haven't used any other scripts, tools or programs that might have caused this issue.

### Windows version

win11 23H2

### Script mode/options

without any options

### Describe the issue

When users download the Win11Debloat ZIP archive directly from GitHub (via browser) and extract it, Windows adds a "Mark of the Web" (Zone.Identifier) to all extracted files. This causes PowerShell to prompt a security warning for every single .ps1 file the script tries to load via dot-sourcing.

GUI depends on loading multiple script files, these prompts interrupt the initialization – the script cannot complete the load and falls back to CLI mode instead of showing the graphical interface.

A simple fix is to run(under the extracted folder):

```powershell
Get-ChildItem -Path . -Recurse -File | Unblock-File
```

This removes the block in one step and allows the GUI to start normally.

Please consider adding this command to the README (under "Usage" or a new "Troubleshooting" section). It would make the tool more beginner-friendly.

Thank you for your work!

### Steps to reproduce

Download the latest Win11Debloat ZIP from the Releases page using a browser.

Extract the ZIP to a folder.

Open PowerShell as Administrator and navigate to the extracted folder.

Run .\Win11Debloat.ps1 (or Run.bat).

### Error output

_No response_

### Additional context

_No response_

HetCreep:
PROBLEM: Downloading the release ZIP with a browser and extracting it with Explorer stamps every file with the Zone.Identifier (Mark-of-the-Web) stream. Under a RemoteSigned-or-stricter effective execution policy, every dot-sourced .ps1 then triggers a per-file security prompt or is blocked outright -- breaking or interrupting the GUI load (#720). Critically, the reporter confirmed (via Get-ExecutionPolicy -List) an Unrestricted policy at MachinePolicy scope on their machine -- and a Group-Policy-scoped policy OVERRIDES the -ExecutionPolicy Bypass that Run.bat passes, so no launcher flag can prevent the prompts. The only fix that works regardless of policy scope is removing the stream itself. FIX: one line at startup (after the PowerShell 7 guard, before the admin check, so it runs before any dot-sourcing in both the non-elevated and relaunched-elevated process): Get-ChildItem -Path PSScriptRoot -Recurse -File piped to Unblock-File with SilentlyContinue. Unblock-File deletes the Zone.Identifier alternate data stream; on unmarked files it is a no-op, and errors (e.g. exotic ACLs) are suppressed so the line can never break startup. VERIFICATION: reproduced the full cycle non-destructively on Windows PowerShell 5.1 (5.1.26100): created a .ps1, stamped it with a real Zone.Identifier stream (ZoneId=3), confirmed dot-sourcing it in a RemoteSigned process is blocked, ran the exact Get-ChildItem/Unblock-File line from this PR, confirmed the stream is gone, and confirmed the same dot-source then loads cleanly. Parse-clean. The reporter on #720 has offered to test this change on the affected machine (GPO-scoped policy), which covers the scenario I cannot reproduce locally. NOTE: this addresses the failure class at the code level; the README suggestion from #720 remains complementary if you also want the docs note. Fixes #720

# Summary by CodeRabbit

- Add startup cleanup for `Zone.Identifier` streams on `.ps1`, `.psm1`, and `.psd1` files.
- Run cleanup only when the main script is marked and `-WhatIf` is not active.
- Suppress cleanup errors so startup continues.
- Preserve Mark-of-the-Web data for non-PowerShell files.

### DIFF
--- a/Win11Debloat.ps1
+++ b/Win11Debloat.ps1
@@ -116,20 +116,27 @@ if ($PSVersionTable.PSEdition -eq 'Core') {
 
 # Remove the Mark-of-the-Web (Zone.Identifier) from the PowerShell script files so they
 # dot-source without per-file security prompts. The stream is added when the ZIP is
-# downloaded with a browser and extracted with Explorer, and a Group-Policy-scoped
-# execution policy overrides the -ExecutionPolicy Bypass that Run.bat passes, so launcher
-# flags alone cannot prevent the prompts. Only run when this file itself is still marked
-# (skips the recursive sweep on every normal launch once the folder is already clean) and
-# never in -WhatIf mode. Only script/module files are unblocked -- other file types are not
-# affected by the mark and keep their zone information. See issue #720.
+# downloaded with a browser and extracted with Explorer, and -ExecutionPolicy Bypass (Run.bat)
+# sets only the Process scope, which the MachinePolicy/UserPolicy scopes (Group Policy) outrank
+# -- so this can only matter when one of those two is actually set; for everyone else Bypass
+# already wins and this whole block, including the cheap self-mark check below, is skipped.
+# Never runs in -WhatIf mode. Only run when this file itself is still marked (skips the
+# recursive sweep on every normal launch once the folder is already clean). Only script/module
+# files are unblocked -- other file types are not affected by the mark and keep their zone
+# information. See issue #720.
 if (-not $WhatIfPreference) {
-    $selfBlocked = [bool](Get-Item -LiteralPath $PSCommandPath -Stream * -ErrorAction SilentlyContinue |
-        Where-Object { $_.Stream -eq 'Zone.Identifier' })
+    $gpoExecutionPolicySet = (Get-ExecutionPolicy -Scope MachinePolicy) -ne 'Undefined' -or
+        (Get-ExecutionPolicy -Scope UserPolicy) -ne 'Undefined'
 
-    if ($selfBlocked) {
-        Get-ChildItem -Path $PSScriptRoot -Recurse -File |
-            Where-Object { $_.Extension -in '.ps1', '.psm1', '.psd1' } |
-            Unblock-File -ErrorAction SilentlyContinue
+    if ($gpoExecutionPolicySet) {
+        $selfBlocked = [bool](Get-Item -LiteralPath $PSCommandPath -Stream * -ErrorAction SilentlyContinue |
+            Where-Object { $_.Stream -eq 'Zone.Identifier' })
+
+        if ($selfBlocked) {
+            Get-ChildItem -Path $PSScriptRoot -Recurse -File |
+                Where-Object { $_.Extension -in '.ps1', '.psm1', '.psd1' } |
+                Unblock-File -ErrorAction SilentlyContinue
+        }
     }
 }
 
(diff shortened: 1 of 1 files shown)


HetCreep:
Good call -- addressed. The sweep is now filtered to .ps1/.psm1/.psd1 only, which matches what the mark actually blocks (dot-sourcing); json/xaml/reg/assets are read through APIs that ignore MOTW, so their zone information now stays intact and any future non-script payload keeps its SmartScreen coverage. Re-verified the filtered pipeline end-to-end on PS 5.1: a marked .ps1 + .json + .reg trio -> only the .ps1 is unblocked (and then dot-sources cleanly under RemoteSigned), the .json and .reg remain marked.

HetCreep:
Confirmed working on the affected machine (GPO-scoped MachinePolicy=Unrestricted) -- from #720: "I've just tested the updated version, and everything works as expected -- the GUI launches without any security warnings, and all features are accessible. No more repeated prompts." That covers the one real-world scenario I couldn't reproduce locally, on top of the local end-to-end proof above.

Raphire:
Heya,

I think this addition is fine, but I don't really like that these changes to files are always made, even when it's not really needed and when `-WhatIf` is passed.

HetCreep:
Fair points, both addressed. Now gated on two conditions: (1) skipped entirely under -WhatIf (WhatIfPreference), since it's a filesystem change; (2) only runs the recursive sweep when the main script itself still carries the Zone.Identifier stream -- a cheap single-file check, so a normal repeat launch (already-clean folder, or a git clone that was never marked) does zero extra filesystem work per run. Verified all three paths on PS 5.1: marked-self -> sweep runs and unblocks children; already-clean-self -> sweep skipped; -WhatIf -> sweep skipped even when marked.

Raphire:
Heya,

I mistakenly committed something here, so I rebased the branch. Sorry about that. 

I think the new changes do improve things, but it still executes needlessly for the vast majority of people who don't have a GPO override on the PowerShell execution policy. With stuff like this I'd rather err on the side of caution and only execute things like this when absolutely necessary.

The machine policy mentioned in #720 should be sufficient, but we should confirm there aren't any other possible settings that could cause issue.

HetCreep:
Heya,

Good point, and confirmed against Microsoft's execution-policy docs: the only scopes that outrank `-ExecutionPolicy Bypass` (Process scope) are **MachinePolicy** and **UserPolicy** (both Group Policy only, per `about_Execution_Policies`) -- there's no third setting that can cause this.

So the check is now gated on that directly: before even touching `` to check the self-mark, it checks whether either GPO scope is set (`Get-ExecutionPolicy -Scope MachinePolicy`/`-Scope UserPolicy`, both `!= 'Undefined'`). Only when one of those is actually set does the self-mark check -- and therefore the sweep -- run at all. For everyone without a GPO override, which is the vast majority, neither the self-check nor the sweep execute anymore.

Verified on a clean machine (no GPO override, both scopes `Undefined`): the gate evaluates `False` and the whole block is skipped. Also mocked the #720 scenario (`MachinePolicy` set) to confirm the gate opens correctly when it should.

Raphire:
Heya,

Thanks for the changes. I made some adjustments to allow logging and cleaned up the comments a bit to keep things readable.

PR #732, opened 2026-08-09
Security hardening: launcher TOCTOU fix and safe force-Edge removal

plwrd:
## Summary

Two security fixes found during a review of the codebase:

### 1. `Scripts/Get.ps1` — unpacked script files were executed elevated from a user-writable directory

The launcher downloaded and unpacked the archive into `%TEMP%\Win11Debloat` and then ran `Win11Debloat.ps1` elevated from there. Between unpacking and elevation (and during the elevated run), any non-elevated process running as the same user could swap the unpacked `.ps1` files - which `Win11Debloat.ps1` dot-sources - and get its own code executed with administrator rights behind the UAC prompt the user intended for Win11Debloat.

The launcher now:
- records the SHA-256 of the archive immediately after download
- elevates a bootstrap that creates (or takes over) `%ProgramData%\Win11Debloat`, restricting it to **Administrators + SYSTEM** with inheritance disabled - this also neutralizes a pre-created attacker directory
- copies the archive into the protected directory and **re-verifies the hash there**, aborting on mismatch
- unpacks, restores `LastUsedSettings.json`, runs `Win11Debloat.ps1` and cleans up entirely inside the protected directory
- migrates existing `Config`, `Logs` and `Backups` from the old `%TEMP%\Win11Debloat` location on first run

Behavior change: the working directory moves from `%TEMP%\Win11Debloat` to `%ProgramData%\Win11Debloat`, which is inherent to the fix - the staging area must be somewhere a non-admin process cannot write.

### 2. `Scripts/AppRemoval/Invoke-ForceRemoveEdge.ps1` — could delete the real EdgeHTML package on Windows 10

`ForceRemoveEdge` unconditionally created a stub at `%SystemRoot%\SystemApps\Microsoft.MicrosoftEdge_8wekyb3d8bbwe` and later recursively deleted that path. On Windows 10 this folder is the genuine legacy Edge (EdgeHTML) system package: `New-Item` failed non-terminally, execution continued, and the real system component was deleted. `ForceRemoveEdge` has no `MinVersion` gate in `Features.json`, so nothing prevented this on Windows 10.

Fixed by:
- locating the uninstaller **before** mutating anything, bailing out early when missing
- only creating the stub when the folder does not already exist, and only deleting it if this run created it
- cleaning up the stub in a `finally` block so no fake `MicrosoftEdge.exe` is left in `SystemApps` when the uninstall fails (previously it was left behind permanently when the uninstall registry key was missing)
- invoking the `UninstallString` directly via `Start-Process` instead of interpolating the raw registry value into a `cmd.exe /c` command line

## Test plan

- Full Pester suite: **450 passed, 0 failed** (`Scripts/Run-Tests.ps1`)
- Both files pass PowerShell parser validation; the embedded elevated bootstrap template was rendered and parser-validated as well

🤖 Generated with [Claude Code](https://claude.com/claude-code)

# Summary by CodeRabbit

- Hardened `Scripts/Get.ps1` against launcher TOCTOU attacks.
- Downloaded, verified, extracted, and executed archives in the elevated `%ProgramData%\Win11Debloat` staging directory.
- Secured staging with ACL checks, reparse-point protection, unique archive names, and concurrent-run serialization.
- Migrated configuration, logs, and backups from `%TEMP%\Win11Debloat`.
- Preserved configuration after failed unpacking and retained staged files when the launched script fails.
- Propagated elevated bootstrap failures to the launcher.
- Escaped literal bootstrap inputs to prevent recursive token-replacement corruption.
- Hardened Edge removal by validating the uninstaller, invoking it directly, and cleaning up conditional stubs with `finally`.
- Added comment-based help for `Format-EmbeddedLiteral`.
- All 450 Pester tests pass.
- PowerShell parser validation passes for the modified files and elevated bootstrap template.

### DIFF
--- a/Scripts/AppRemoval/Invoke-ForceRemoveEdge.ps1
+++ b/Scripts/AppRemoval/Invoke-ForceRemoveEdge.ps1
@@ -49,10 +49,18 @@ function Invoke-ForceRemoveEdge {
 
         $uninstallArgs = ($uninstallArgs + ' --force-uninstall').Trim()
 
-        Invoke-NonBlocking -ScriptBlock {
+        $uninstallExitCode = Invoke-NonBlocking -ScriptBlock {
             param($exe, $arguments)
-            Start-Process -FilePath $exe -ArgumentList $arguments -WindowStyle Hidden -Wait
+            $uninstallProcess = Start-Process -FilePath $exe -ArgumentList $arguments -WindowStyle Hidden -Wait -PassThru
+            return $uninstallProcess.ExitCode
         } -ArgumentList $uninstallExe, $uninstallArgs
+
+        # Don't clean up shortcuts and autostart entries when the uninstaller failed,
+        # Edge is still installed in that case
+        if ($null -ne $uninstallExitCode -and $uninstallExitCode -ne 0) {
+            Write-Host "Unable to forcefully uninstall Microsoft Edge, the uninstaller failed with exit code $uninstallExitCode" -ForegroundColor Red
+            return
+        }
 
         Write-Host "Removing leftover files..."
 
--- a/Scripts/AppRemoval/Invoke-ForceRemoveEdge.ps1
+++ b/Scripts/AppRemoval/Invoke-ForceRemoveEdge.ps1
@@ -7,22 +7,52 @@ function Invoke-ForceRemoveEdge {
 
     $regView = [Microsoft.Win32.RegistryView]::Registry32
     $hklm = [Microsoft.Win32.RegistryKey]::OpenBaseKey([Microsoft.Win32.RegistryHive]::LocalMachine, $regView)
+
+    # Locate the uninstaller first, before making any changes to the system
+    $uninstallRegKey = $hklm.OpenSubKey('SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\Microsoft Edge')
+    $uninstallString = if ($null -ne $uninstallRegKey) { [string]$uninstallRegKey.GetValue('UninstallString') } else { $null }
+
+    if ([string]::IsNullOrWhiteSpace($uninstallString)) {
+        Write-Host "Unable to forcefully uninstall Microsoft Edge, uninstaller could not be found" -ForegroundColor Red
+        return
+    }
+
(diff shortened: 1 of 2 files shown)


plwrd:
Thanks for the reviews — pushed fixes for all findings:

- **Bootstrap exit code** now captured after `WaitForExit()` and propagated, so download/integrity/extraction/script failures no longer report success.
- **Pre-elevation hash removed** — correct that it wasn't a real check; an attacker who can swap the `%TEMP%` file is present before the hash runs. The download now happens inside the elevated bootstrap, directly into the Administrators/SYSTEM-only staging directory over TLS, so there's no user-writable intermediate to swap.
- **Token substitution** replaced with a header of escaped single-quoted literals, so an input containing a token string can't be corrupted by a later pass.
- **Staging TOCTOU** — reject/delete reparse points, refuse to adopt a directory not already owned by Administrators/SYSTEM, and re-verify ownership + reparse status *after* the admin-only ACL is applied (once that ACL is in place a non-elevated process can't swap the object).

On opening the directory via a native handle with `FILE_FLAG_OPEN_REPARSE_POINT`: that needs P/Invoke, which I've kept out of this launcher since it ships as a plain bundled `.ps1`. With the download now landing directly in the owner-restricted directory plus the owner-check and post-ACL re-verification, I believe the residual window is closed for a non-elevated attacker in practice — happy to revisit if you'd prefer the native-handle approach.

Raphire:
Heya,

Thanks for taking the time to contribute! I haven't had a chance to test or fully review the code yet, but I do have some initial thoughts.

1. While I can see the potential security issue here with a targeted attack replacing script files mid-run, I don't really like the current approach. It changes the behaviour quite considerably by saving everything to a shared directory and using a global mutex to guard against potential issues. The string encoded bootstrap function is also problematic as it's difficult to read, debug and extend in the future. 

In the current state this adds a lot of complexity and introduces a lot of potential issues. 

2. Overall, I like he improvements to the edge removal script. I do see a potential issue with Start-Process returning $null and being treated as a success. Also, the $edgeStub existence check doesn't explicitly check for a folder.

<hr>

Before we continue with this, would you mind splitting these into two separate PRs? Keeping each PR focused on a single change makes reviewing much easier and lets us merge them independently.

PR #737, opened 2026-08-13
fix: match tweak presets and app-removal scope by stable ID instead of translatable text

SashankBhamidi:
Prerequisite for the localization work in #499. Get-CategoryTweakPresetMap and Update-AppRemovalScopeDescription both matched against display text (category Name, ComboBoxItem Content) instead of a stable identifier. Change either one, translation or not, and both silently stop matching.

Added CategoryId to Categories in Features.json, threaded it through UiControlMappings, and switched both functions to match on stable IDs (CategoryId, ComboBoxItem Name) instead.

# Summary by CodeRabbit

- Added stable `CategoryId` values to all category definitions.
- Stored `CategoryId` in feature and group UI mappings, with legacy fallback support.
- Updated `Get-CategoryTweakPresetMap` to match by `CategoryId`.
- Updated app-removal scope handling to use stable `ComboBoxItem.Name` values.
- Added target resolution and username validation for app removal.
- Added tests for category fallback, translated scope descriptions, target resolution, and username validation.

### DIFF
--- a/Config/Features.json
+++ b/Config/Features.json
@@ -2,50 +2,62 @@
   "Version": "1.0",
   "Categories": [
     {
+      "CategoryId": "PrivacySuggestedContent",
       "Name": "Privacy & Suggested Content",
       "Icon": "&#xE72E;"
     },
     {
+      "CategoryId": "System",
       "Name": "System",
       "Icon": "&#xe770;"
     },
     {
+      "CategoryId": "StartMenuSearch",
       "Name": "Start Menu & Search",
       "Icon": "&#xe8fc;"
     },
     {
+      "CategoryId": "AI",
       "Name": "AI",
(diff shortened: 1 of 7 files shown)


Raphire:
Heya,

The changes look generally fine, I noticed just 1 issue.

In Show-MainWindow.ps1 (line 677) the scope is still added based on the combo Content, instead of the Name. This will cause the AppRemovalScope to fall back to AllUsers if it gets translated.

EDIT: Not really a blocker, but it's good to note that the current approach still uses the Category name in Feature definitions. This probably works just fine, but it may be worth looking into migrating this to just using Category id's wherever possible.

Raphire:
Hey, thanks for the changes! I'll merge this in.

PR #745, opened 2026-08-20
Surface the underlying GitHub error when the launcher download fails

ryanduguid:
## Summary
- When `Scripts/Get.ps1` cannot download Win11Debloat from GitHub, keep the original connection error and append the underlying exception.
- Also append GitHub `ErrorDetails` when they add information, such as API rate-limit messages.
- Tests parse `Get.ps1` and evaluate only the new helper, so the launcher is never executed.

Closes https://github.com/Raphire/Win11Debloat/discussions/743

## Test plan
- [ ] `Tests/GetLauncher.Tests.ps1` covers exception text and distinct `ErrorDetails`
- [ ] Confirm a simulated 403/rate-limit failure prints both the original message and the GitHub reason

# Summary by CodeRabbit

- Added `Get-GitHubDownloadFailureMessage` to preserve the original connection error and append the underlying exception.
- Included distinct GitHub `ErrorDetails`, including API rate-limit messages.
- Updated the download failure handler to use the new helper.
- Added isolated Pester tests that parse `Get.ps1` and validate the helper behavior.

### DIFF
--- a/Scripts/Get.ps1
+++ b/Scripts/Get.ps1
@@ -105,6 +105,23 @@ param (
     [switch]$HideDriveLetters
 )
 
+<#
+    .SYNOPSIS
+        Formats the user-facing error shown when the launcher cannot download Win11Debloat.
+
+    .DESCRIPTION
+        Always starts with the generic GitHub download failure message. Appends the
+        exception message when it is present, then appends ErrorDetails.Message when
+        it is present and different from the exception text (for example a GitHub
+        API rate-limit body).
+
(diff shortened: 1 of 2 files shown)


Raphire:
Heya,

Thanks for taking the time to contribute. While your addition seems fine functionally, I think it's a bit too overly complicated. I have opted for a simpler approach to accomplish more verbose error output here: https://github.com/Raphire/Win11Debloat/pull/753

PR #764, opened 2026-09-04
Add localization framework for the GUI (en-US baseline)

SashankBhamidi:
Related to #499.

The GUI's been English-only since day one, and #499 has people asking for this pretty regularly. This is the framework underneath it, plus the English content everything else gets built and checked against. No second language yet, that's deliberate, get the base right first so a translation doesn't inherit shortcuts.

There was an earlier attempt at this in #643. I read through it before starting and kept the marker-substitution idea, but rebuilt the lookup itself, since the original had five near-identical functions doing the same job with different property names, keyed categories by their English name so renaming one silently broke its translation, and had no tests on any of it. Nothing from that branch carries over directly, this is a clean implementation, but the approach working at all is owed to it.

`Get-Translation -Key` for PowerShell, `%LANG:Key%` for XAML, one lookup instead of five. Falls back to en-US at every level, a missing folder, a missing key, even a bad plural rule, so it degrades to English instead of crashing or showing a raw key on screen. Plural strings use a `_one`/`_other` suffix instead of hardcoded English grammar. `Test-LanguageKeyCoverage` diffs a language against en-US so you can tell if a translation's actually finished instead of guessing. `-Language` on `Win11Debloat.ps1` lets you preview a language without touching your Windows display settings.

CLI stays English for now, that's separable and bigger on its own.

# Summary by CodeRabbit

- Added `en-US` GUI localization resources.
- Added translation lookup, fallback, pluralization, formatting, coverage checks, and XAML marker conversion.
- Replaced hardcoded GUI and script messages with localized lookups.
- Added the `-Language` parameter for language preview.
- Added UTF-8 handling for BOM-free JSON files.
- Added localization tests and contributor documentation.
- The CLI remains English-only.


yasar11732:
I had already started translating to Turkish based on earliear PR. Should I stop working on it until one of the PRs get merged?

Raphire:
> I had already started translating to Turkish based on earliear PR. Should I stop working on it until one of the PRs get merged?

Thanks for taking the time to translate. I would hold off until we get this merged in, there are quite a few differences between this and the previous PR when it comes to translation keys, etc.

Raphire:
I haven't had the time to fully review it yet, I'll try to do that this weekend or the coming week. 

I like the general approach, but I did notice that the apps are not translated. I think this would be good to include here due to the descriptions and localized app names that people would expect to see.

SashankBhamidi:
Good call, and it's the natural next piece. Apps.json fits the same schema, keyed by AppId, same Get-Translation lookup, same fallback and coverage-check behavior everything else in this PR already has. Didn't fold it in here so the diff stays focused on the framework itself and easy to review.

I'll open it as a follow-up right after this lands, before any translation work starts on top of it, so translators aren't hitting a moving target for keys.

Raphire:
I've had some more time to review this. I like the general approach here, and I think the PR is in a good place. 

I did notice however that some text will not be shown correctly. E.g. Español will be shown as EspaÃ±ol when imported. This is probably a text encoding issue that doesn't surface for the English translations because that is ASCII only.

Other than that, I don't see any blocking issues 😁

SashankBhamidi:
@Raphire,

Hmm. Good catch. `Get-Content` was falling back to the system codepage for files without a BOM, mangling non-ASCII text. 

Not Spanish-specific, it's in the shared loader every language file goes through, so it'd hit any accented/non-ASCII characters. 

Forced UTF-8 explicitly, fixed in fa760d2.

Used AI for `Config.Utf8NoBom.json`.

Raphire:
> @Raphire,
> 
> Hmm. Good catch. `Get-Content` was falling back to the system codepage for files without a BOM, mangling non-ASCII text.
> 
> Not Spanish-specific, it's in the shared loader every language file goes through, so it'd hit any accented/non-ASCII characters.
> 
> Forced UTF-8 explicitly, fixed in [fa760d2](https://github.com/Raphire/Win11Debloat/commit/fa760d2dd9a5d719a327e1e81f4dea9c17637441).
> 
> Used AI for `Config.Utf8NoBom.json`.

Thanks for the changes! I don't see any other issues so I'll merge it in so I, @yasar11732 and others can start with the translations.

